PR #2, opened 2018-09-21
Tickets/dm 15300

menanteau:


ktlim:
You may want to consider https://developer.lsst.io/work/flow.html#appendix-commit-organization-best-practices
`git add -p` can be very useful to separate out conceptually different modifications into distinct commits.  Any time your commit message says "in addition" or "and", it might be an indication that this should be two commits.  (But we all combine at times.)

You might also consider https://developer.lsst.io/work/flow.html#appendix-commit-message-best-practices
In particular, use of the imperative and 50 character initial lines seem to be missing here.

ktlim:
You should never be merging master into a tickets branch.  You should be rebasing instead.

menanteau:
That was not the intention, I though that I was doing the opposite!

bin/DMHS_ATS_configurable
```diff
@@ -31,29 +30,30 @@ def cmdline():
...
-    parser.add_argument("--send_efd_message", action="store_true",default=False, 
+    parser.add_argument("--send_efd_message", action="store_true", default=False,
                         help="Send and EDF message as well?")
-    #args = parser.parse_args()
+    # args = parser.parse_args()
     # Set the defaults of argparse using the values in the yaml config file
     parser.set_defaults(**conf_defaults)
     args = parser.parse_args(args=remaining_argv)
 
-    #print ("Will use:")
```
ktlim, line 53: Debugging code can often be left in, rather than commented out, with output only at DEBUG level.

Also, we typically do not use a space between `print` and the parenthesis, as `print` is a function call.  This needs to be fixed throughout.

bin/LFO_dmHeaderService_EventLogger
```diff
@@ -26,7 +25,7 @@ if __name__ == "__main__":
             LOGGER.info("ID: %s" % event.id)
             LOGGER.info("Version: %s" % event.version)
             LOGGER.info("priority: %s" % event.priority)
-            retval==-1
+            retval == -1
```
ktlim: `==` is probably not what is desired.

bin/assemble_FITS
```diff
@@ -35,60 +35,63 @@ def cmdline():
...
+    outname = kwargs.get('outname', None)
+    byext = kwargs.get('byext', False)
+    dtype = kwargs.get('dtype', numpy.int32)
+    compress_type = kwargs.get('compress_type', None)
 
     # Make sure that we have sane values for NAXIS1/NAXIS1
     if naxis1 == 0 or naxis2 == 0:
-        print("WARNING: Both NAXIS1/NAXIS2 muste be non-zero, current values are:{},{}".format(naxis1,naxis2))
-    
+        print("WARNING: Both NAXIS1/NAXIS2 muste be non-zero, \
```
ktlim: Backslash continuation within a quoted string is unusual and probably does not have the effect desired (particularly with regard to spacing).  Concatenation is more typical.

python/HeaderService/camera_coords.py
```diff
@@ -1,6 +1,10 @@
 import HeaderService
+import math
+
+"""This modules describe the CCD/Camera coordinate systems as described in
```
ktlim: Here and elsewhere, Python files need to have the LSST license header, and docstrings should conform to the standards in https://developer.lsst.io/python/numpydoc.html

bin/assemble_FITS
menanteau: Fixed

bin/DMHS_ATS_configurable
menanteau: Cleaned this up.

bin/LFO_dmHeaderService_EventLogger
menanteau: Changed to `retval = -1`

python/HeaderService/camera_coords.py
menanteau: If understand this correctly, the string is correctly formatted, but I'm missing the License? If that is the case, I'd like to create a new separate ticket to add the License to all relevant files in both salpytools and headerService.

timj: "This module" not "This modules"

menanteau: Typo had been fixed too. Thanks for being an attentive reader.

menanteau: I've created a ticket to add the LSST license header as a separate task, DM-15861.

PR #3, opened 2018-10-08
Add logging level options.

menanteau:
Different logging levels can be set from the command using the --log option and from the input yaml configuration file

### DIFF
--- a/bin/DMHS_ATS_configurable
+++ b/bin/DMHS_ATS_configurable
@@ -2,15 +2,9 @@
 
 import os
 import logging
-import HeaderService.hutils as hutils
 import HeaderService.hslib as hslib
 import argparse
 import yaml
-
-spinner = hutils.spinner
-
-# Create a logger for all functions
-LOGGER = hutils.create_logger(level=logging.NOTSET, name='HEADERSERVICE')
 
 
 def cmdline():
@@ -45,10 +39,15 @@ def cmdline():
                         help="Port number for webserver")
     parser.add_argument("--send_efd_message", action="store_true", default=False,
                         help="Send and EDF message as well?")
+    parser.add_argument("--loglevel", action="store", default='INFO', type=str.upper,
+                        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+                        help="Logging Level [DEBUG/INFO/WARNING/ERROR/CRITICAL]")
+
     # args = parser.parse_args()
     # Set the defaults of argparse using the values in the yaml config file
     parser.set_defaults(**conf_defaults)
     args = parser.parse_args(args=remaining_argv)
+    args.loglevel = getattr(logging, args.loglevel)
     return args
 
 
--- a/etc/conf/atTelemetry.yaml
+++ b/etc/conf/atTelemetry.yaml
@@ -1,5 +1,8 @@
 # Configuration file for the AuxTelescope
 
+# Default logging level
+log: INFO
+
 # The camera vendor
 vendor: ITL
 
@@ -17,7 +20,7 @@ imageParam_event:
 # The URL format to be used
 url_format: "http://{ip_address}:{port_number}/{filename_HDR}"
 
-# The file naming event source and definitions 
+# The file naming event source and definitions
 filepath: 'www/DMHS_filerepo'
 format_HDR: "{}.header"
 format_FITS: "{}.fits"
@@ -26,8 +29,8 @@ imageName_event:
     topic: startIntegration
     value: imageName
 
-# The Events that we use to define where we collect 
-start_collection_event: 
+# The Events that we use to define where we collect
+start_collection_event:
  device: atcamera
  topic: startIntegration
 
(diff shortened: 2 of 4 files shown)


bin/DMHS_ATS_configurable
```diff
@@ -45,13 +39,17 @@ def cmdline():
                         help="Port number for webserver")
     parser.add_argument("--send_efd_message", action="store_true", default=False,
                         help="Send and EDF message as well?")
+    parser.add_argument("--log", action="store", default='INFO',
+                         choices=["DEBUG","INFO","WARNING","ERROR","CRITICAL"],
```
timj: There is a flake8 violation here since you need a space after the comma.  Please set up continuous integration for PRs so that flake8 is automatically run.  I can help you with this.

bin/DMHS_ATS_configurable
```diff
@@ -45,13 +39,17 @@ def cmdline():
                         help="Port number for webserver")
     parser.add_argument("--send_efd_message", action="store_true", default=False,
                         help="Send and EDF message as well?")
+    parser.add_argument("--log", action="store", default='INFO',
```
timj: I think I would prefer `--loglevel` or `--log-level` as the name of this parameter. That's what the parameter becomes internally.

bin/DMHS_ATS_configurable
menanteau: Thanks for catching that. Yes, I'd like to setup continuous integration.

bin/DMHS_ATS_configurable
```diff
@@ -45,13 +39,17 @@ def cmdline():
...
+                         choices=["DEBUG","INFO","WARNING","ERROR","CRITICAL"],
+                         help="Logging Level [DEBUG/INFO/WARNING/ERROR/CRITICAL]")
+
     # args = parser.parse_args()
     # Set the defaults of argparse using the values in the yaml config file
     parser.set_defaults(**conf_defaults)
     args = parser.parse_args(args=remaining_argv)
+    args.loglevel = getattr(logging, args.log.upper())
```
timj: The upper casing here is not doing anything because the choices above mandates that only upper case options are allowed. I think you need to add `type=str.upper` to your `add_argument()` call above so that lower case `debug` is converted to `DEBUG` before the choices are checked.

etc/conf/atTelemetry.yaml
```diff
@@ -1,5 +1,8 @@
 # Configuration file for the AuxTelescope
 
+# logging level
```
timj: Can you add `Default` here so that it reads `Default logging level`?

bin/DMHS_ATS_configurable
menanteau: `--loglevel` was my original parameter, but the changed it to `--log `because it looked simpler. Happy to change it `--loglevel`

bin/DMHS_ATS_configurable
```diff
@@ -45,13 +39,17 @@ def cmdline():
...
     return args
 
-
```
timj: Please put whitespace changes in a separate commit.

bin/DMHS_ATS_configurable
timj: Sorry. I was doing a review but got distracted by a side conversation and ended up pushing "Add Single comment" all the time.

python/HeaderService/hutils.py
```diff
@@ -358,7 +358,7 @@ def update_records(self, newdict, extname):
                 self.update_record(keyword, value, extname)
                 LOGGER.debug("Updating {}".format(keyword))
             except Exception:
-                LOGGER.debug("WARNING: Could not update {}".format(keyword))
+                LOGGER.warning("WARNING: Could not update {}".format(keyword))
```
timj: I would prefer that the changes of warning levels are made in a separate commit distinct from the commit that adds the ability to override the log level.

bin/DMHS_ATS_configurable
timj: loglevel is consistent with what science pipelines are doing (pytest uses `log-level` so there isn't consistency for python).

etc/conf/atTelemetry.yaml
menanteau: Yes. Done.

python/HeaderService/hutils.py
menanteau: Reverted to debug.

timj: It looks like `logging` suggests `.exception()` be used for logging exception information.

timj:
@menanteau I had hoped you would be able to split the commit into separate commits: whitespace, flake8, addition of -loglevel, but I see the pull request has been merged so it is too late to fix now. `git add -p` is a very useful command that can be used to add subsets of changes to the index before committing. This lets you make unrelated changes but commit them as distinct commits. Slack me if you want more information on how to do this.

menanteau:
@timj I'm sorry, but I misinterpreted your suggestion of putting the whitespace, etc in a different commit as a not belonging to the current branch. Reading it now it makes total sense as a different commit within the *same* branch. I get it now. But I guess is too late to fix it. Let me know if that is not the case.

PR #5, opened 2018-10-15
DM-16053: Update write_header function

menanteau:


python/HeaderService/hslib.py
```diff
@@ -21,6 +21,7 @@
 # Create a logger for all functions
 LOGGER = hutils.create_logger(level=logging.NOTSET, name='HEADERSERVICE')
 
+
```
timj, line 24: Whitespace/flake8 fixes should be in separate commits from functionality changes.

python/HeaderService/hutils.py
```diff
@@ -384,17 +275,23 @@ def write_headers(self, filenames, MP=False, NP=2, md5=False):
...
+    def write_header_string(self, filename):
+        """ Write header as string"""
+        self.string_header()
+        with open(filename, 'w') as fobj:
+            fobj.write(self.hstring)
+
     def write_fits(self, filename, dtype='random', naxis1=None, naxis2=None, btype='int32'):
 
-        """ Write a dummy fits file -- use for testing only"""
+        """ Write a dummy fits file filles with random or zeros -- use for testing only"""
```
timj: "filled" rather than "filles"

python/HeaderService/hutils.py
```diff
@@ -384,17 +275,23 @@ def write_headers(self, filenames, MP=False, NP=2, md5=False):
...
                 hdr = copy.deepcopy(self.header[extname])
                 fits.write(data, header=hdr, extname=extname)
 
+    def write_header_string(self, filename):
+        """ Write header as string"""
+        self.string_header()
+        with open(filename, 'w') as fobj:
+            fobj.write(self.hstring)
+
     def write_fits(self, filename, dtype='random', naxis1=None, naxis2=None, btype='int32'):
```
timj: Maybe clearer to name it as `write_dummy_fits` ?

python/HeaderService/hutils.py
```diff
@@ -384,17 +275,23 @@ def write_headers(self, filenames, MP=False, NP=2, md5=False):
...
+        """Write a header file using the FITS format with empty HDUs"""
         data = None
         with fitsio.FITS(filename, 'rw', clobber=True, ignore_empty=True) as fits:
             for extname in self.HDRLIST:
                 hdr = copy.deepcopy(self.header[extname])
                 fits.write(data, header=hdr, extname=extname)
 
+    def write_header_string(self, filename):
+        """ Write header as string"""
+        self.string_header()
```
timj: It would look less like magic if the return value of this method was then printed out. `string_header()` would feel less magical I think if it was called `calculate_string_header()` so that it would more explicitly be updating `self`.  In this case it is much clearer to use the return value and not rely on the magic.

python/HeaderService/hutils.py
```diff
@@ -384,17 +275,23 @@ def write_headers(self, filenames, MP=False, NP=2, md5=False):
...
+        with open(filename, 'w') as fobj:
+            fobj.write(self.hstring)
```
timj: I have a slight preference for `print(self.hstring, file=fobj, end='')` as being more expressive, but `write()` is allowed.

python/HeaderService/hutils.py
```diff
@@ -427,19 +322,20 @@ def write_fits(self, filename, dtype='random', naxis1=None, naxis2=None, btype='
...
+        t0 = time.time()
+        if self.write_mode == 'fits':
+            self.write_header_fits(filename)
+        elif self.write_mode == 'string':
+            self.write_header_string(filename)
         else:
-            t0 = time.time()
-            self.write_header_emptyHDU(filename)
-            LOGGER.info("Header write time:{}".format(elapsed_time(t0)))
+            LOGGER.info("ERROR: header write_mode:{} not recognized".format(self.write_mode))
```
timj: This seems like more of a `LOGGER.error` or `LOGGER.warning` rather than `LOGGER.info`.  Given this should never happen and given that you want a header written should it not raise?

timj (approved):
I have some minor comments. Also please start to use numpydoc docstrings as you add/modify methods.

python/HeaderService/hslib.py
menanteau: Will separate these commits.

python/HeaderService/hutils.py
menanteau: fixed

python/HeaderService/hutils.py
menanteau: Changes to `write_dummy_fits` as suggested.

python/HeaderService/hutils.py
menanteau: Yes, it should be `LOGGER.error(message)` followed by a `raise ValueError(msg)`

python/HeaderService/hutils.py
menanteau: Changed to function name to `calculate_string_header()` and use the return value rather than the `self` for clarity.

python/HeaderService/hutils.py
menanteau: will keep the `write()` method as I prefer it.

PR #7, opened 2018-11-02
DM-16222: Update header standards for AuxTel

menanteau:


etc/ATSCam/primary_hdu.header
```diff
@@ -10,14 +10,20 @@ MJD-OBS =                      / Modified Julian Date of observation
...
+OBSGEO-Z=          -3195172.08 / [m] Z-axis Geocentric coordinate
 OBS-NITE=                      / Night of the observation
 TELESCOP= 'LSST AuxTelescope'  / Telescope name
 INSTRUME= 'LATISS'             / Instrument used to obtain these data
 RATEL   =                      / Telescope RA of observation (hr)
 DECTEL  =                      / Telescope DEC of observation (hr)
 RA      =                      / RA of Target (hr)
 DEC     =                      / DEC of Target (deg)
-ROTANGLE=                      / Rotation angle (deg)
+ROTANGLE=                      / Rotation angle relative to the sky (deg)
```
timj, line 34: It would help to have a ROTCOORD header as well that had "sky" or something on it.

etc/ATSCam/primary_hdu.header
```diff
@@ -10,14 +10,20 @@ MJD-OBS =                      / Modified Julian Date of observation
 OBS-LONG=           -70.749417 / [deg] Observatory east longitude
 OBS-LAT =           -30.244639 / [deg] Observatory latitude
 OBS-ELEV=               2663.0 / [m] Observatory elevation
+OBS-LONG=           -70.749417 / [deg] Observatory east longitude
```
timj, line 22: These three are repeated. Are they the AuxTel or LSST locations?

etc/ATSCam/primary_hdu.header
```diff
@@ -27,13 +33,16 @@ OBSERVER=                      / Observer name(s)
...
+CCD_MANU= 'ITL'                / CCD Manufacturer
+CCD_TYPE= '3800C'              / CCD Model Number
+CCD_SERN= '20304'              / Manufacturers? CCD Serial Number
 LSST_NUM= 'ITL-3800C-098'      / LSST Assigned CCD Number
-RAFTNAME= 'LCA-11021_RTM-004_ETU2-Dev' / Name of the RTM
 SEQNAME =                      / SequenceName from Camera StarIntergration
-IMGNAME =                      / ImageName from Camera StarIntergration
+REBNAME =                      / Name of the REB
+IMAGETAG=                      / DAQ Image id
+OBS-ID  =                      / ImageName from Camera StarIntergration
```
timj: "Start" not "Star"

etc/ATSCam/primary_hdu.header
```diff
@@ -10,14 +10,20 @@ MJD-OBS =                      / Modified Julian Date of observation
...
+OBS-LAT =           -30.244639 / [deg] Observatory latitude
+OBS-ELEV=               2663.0 / [m] Observatory elevation
+OBSGEO-X=           1818938.94 / [m] X-axis Geocentric coordinate
+OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
+OBSGEO-Z=          -3195172.08 / [m] Z-axis Geocentric coordinate
 OBS-NITE=                      / Night of the observation
 TELESCOP= 'LSST AuxTelescope'  / Telescope name
 INSTRUME= 'LATISS'             / Instrument used to obtain these data
 RATEL   =                      / Telescope RA of observation (hr)
 DECTEL  =                      / Telescope DEC of observation (hr)
```
timj: Not hours.  This is presumably the tracking coordinate of the bore sight.

etc/ATSCam/primary_hdu.header
```diff
@@ -10,14 +10,20 @@ MJD-OBS =                      / Modified Julian Date of observation
...
 RA      =                      / RA of Target (hr)
 DEC     =                      / DEC of Target (deg)
-ROTANGLE=                      / Rotation angle (deg)
+ROTANGLE=                      / Rotation angle relative to the sky (deg)
 HA      =                      / [HH:MM:SS] Telescope hour angle
 ZD      =                      / [deg] Telescope zenith distance
 AZ      =                      / [deg] Telescope azimuth angle
```
timj: These should state explicitly what time they refer to. "at start" is reasonable. I would actually prefer we had HASTART, ZDSTART, AZSTART and HAEND, ZDEND, ZDEND versions corresponding to DATE-END. Similarly AMSTART, AMEND for airmass.

etc/ATSCam/primary_hdu.header
```diff
@@ -10,14 +10,20 @@ MJD-OBS =                      / Modified Julian Date of observation
...
 OBS-NITE=                      / Night of the observation
 TELESCOP= 'LSST AuxTelescope'  / Telescope name
 INSTRUME= 'LATISS'             / Instrument used to obtain these data
 RATEL   =                      / Telescope RA of observation (hr)
 DECTEL  =                      / Telescope DEC of observation (hr)
 RA      =                      / RA of Target (hr)
 DEC     =                      / DEC of Target (deg)
-ROTANGLE=                      / Rotation angle (deg)
+ROTANGLE=                      / Rotation angle relative to the sky (deg)
 HA      =                      / [HH:MM:SS] Telescope hour angle
```
timj: I don't think it's the telescope hour angle is it? It's the hour angle for the coordinates we are tracking at the telescope location.

etc/ATSCam/primary_hdu.header
```diff
@@ -27,13 +33,16 @@ OBSERVER=                      / Observer name(s)
 PROPOSER=                      / Proposal Principle Investigator
```
timj: Remove PROPOSER

etc/ATSCam/primary_hdu.header
```diff
@@ -27,13 +33,16 @@ OBSERVER=                      / Observer name(s)
...
 PROPID  =                      / Proposal ID
```
timj: I don't think we need PROGRAM and PROPID do we?  Or if we do they should be grouped together in the header and we should be clear how normal survey and deep drilling map to PROPID. Does AuxTel have any concept of program? (There is a typo in the comment for PROGRAM)

timj (commented):
There needs to be a DATE-END for the end of the observation. I have quite a few suggestions so not marking as approved just yet.

Please group related items together rather than scattering them about in the file.  Also, things would be much clearer if you used separators.  For example here's a chunk from a header from my previous telescope:

```
FOCPOSN =                      / [mm] Relative SMU position

        ---- Miscellaneous ----
OCSCFG  = 'scuba2_20091214_053731_962583.xml' / OCS config filename
```
The blank line and the implicit comment really help when scanning through the header.

etc/ATSCam/primary_hdu.header
```diff
@@ -27,13 +33,16 @@ OBSERVER=                      / Observer name(s)
...
 SEQFILE =                      / Name of Sequencer File
```
timj: Do we know what this really means?  @r-owen might know what we would like in here.

timj:
Did I miss TELCODE, DAYOBS, and SEQNUM?

etc/ATSCam/primary_hdu.header
menanteau: Yes, we could add `ROTCOORD` to the header templates, but we'll also need to identify where this will be coming. I could add to the templates for now, but it won't be populated until the proper messages/event is present in SAL. Let me know if this is not OK.

etc/ATSCam/primary_hdu.header
menanteau: Thanks for noticing the duplication. These reflect the LSST location, do you know where I could get the AuxTel exact location?

etc/ATSCam/primary_hdu.header
menanteau: Done

etc/ATSCam/primary_hdu.header
menanteau: I'm not sure if weed PROGRAM or PROPID either. I prefer to remove them now. And if need in the future add them if they are really needed.

etc/ATSCam/primary_hdu.header
menanteau: This is sequencer file used by camera as part of its configuration. Previous chatter on #lsst-ats suggests that it will be need/wanted.

etc/ATSCam/primary_hdu.header
menanteau: Fixed.

etc/ATSCam/primary_hdu.header
menanteau: Removed `(hr) from `DECTEL` and also from `RATEL` until units are better defined.

etc/ATSCam/primary_hdu.header
menanteau: It is unclear at this point the rate at which we will be able to sample `HA`, `ZD` and `AZ` via SAL messages. For now I'll add both `START` and `END` with the optimistic view that we will be able to get them.

etc/ATSCam/primary_hdu.header
menanteau: This is the rotation needed to get a WCS solution. Sometimes this is called `CROTA2`, and it is the coordinate system rotation angle. Maybe we should find a better name for it. Any suggestions?

menanteau:
No you didn't miss `TELCODE`, `DAYOBS` or `SEQNUM`.  I missed them from the conversation from TSEIA-79, but I added them back.

etc/ATSCam/primary_hdu.header
timj: I think if you know that the camera is always publishing it in the same coordinates then it's still better to have the separate keyword. If the camera system can change then we insist that this becomes something that can be read from the system.

etc/ATSCam/primary_hdu.header
timj: HA is the hour angle, not the parallactic angle.  ROTANGLE is the angle of the focal plane relative to RA/Dec if ROOTCOORD is SKY.

etc/ATSCam/primary_hdu.header
```diff
@@ -5,35 +5,52 @@ EXTEND  =                    T / Extensions are permitted
...
+TELCODE = 'AT'                 / The code for the telecope
+CONTRLLR=                      / The controller (e.g. O for OCS, C for CCS)
+DAYOBS  =                      / The observation day as defined in the image name
+SEQNUM  =                      / The sequence number from the image name
+IMAGETAG=                      / DAQ Image id
+SEQNAME =                      / SequenceName from Camera StartIntergration
+REBNAME =                      / Name of the REB
+IMGTYPE =                      / BIAS, DARK, FLAT, OBJECT
+RATEL   =                      / Telescope RA of observation
+DECTEL  =                      / Telescope DEC of observation
```
timj, line 31: Are these in degrees? Also, if these are ICRS then say that in the comment or use RADESYS header.  It could be that these are the apparent RA/Dec rather than the tracking RA/Dec. Which is it?

etc/ATSCam/primary_hdu.header
menanteau: Unfortunately, nothing has been defined by camera or by telescope and site yet. The camera won't publish any coordinates as far as I know. These would only come for T&S. I created a few tickets to make these request a while back.

etc/ATSCam/primary_hdu.header
menanteau: OK -- so I'll include both `ROTANGLE` and `ROTCOORD` and except hope that these will be published by T&S in the future.

etc/ATSCam/primary_hdu.header
menanteau: Unfortunately, the units and the source of these are yet undefined.

menanteau:
@timj I've update the method that reads in the header templates to treat the comments as described in Pence et al. (2010).

PR #12, opened 2019-02-04
Tickets/dm 16444

menanteau:


bin/telemetry_sim_ATS_1VisitNight
```diff
@@ -125,20 +125,24 @@ if __name__ == "__main__":
         time.sleep(args.exptime)
 
         # Send startReadout Event after integration is finished
+        kw['timeStamp'] = time.time()
```
ktlim, line 128: I think we're supposed to use the time from SAL (which will be TAI), not from the system, but as I understand it this is just a simulation for testing so that's not a big problem.

python/HeaderService/hslib.py
```diff
@@ -275,6 +282,16 @@ def run_loop(self):
...
 
+    def timeoutError(self):
+        # Send error code for timeout case
+        kw = {'errorCode': 3010,
+              'errorReport': "Timeout while waiting for {} Event".format(self.name_end),
+              'traceback': '',
+              'priority': 1,
+              }
+        self.dmhs.send_Event('errorCode', **kw)
+        LOGGER.info("Sent erroCode: {}".format(kw))
```
ktlim: `erroCode` → `errorCode`

python/HeaderService/hslib.py
```diff
@@ -265,6 +267,11 @@ def run_loop(self):
                     self.announce()
                     self.EndTelem.newEvent = False
                     LOGGER.info("------------------------------------------")
+                elif self.EndTelem.timeoutEvent:
+                    sys.stdout.flush()
+                    LOGGER.warning("Received timeout: {} Signal".format(self.name_end))
+                    LOGGER.info("Sending Error Event for timeout")
+                    self.timeoutError()
```
ktlim, line 274: Seems a bit dangerous to have the `!newEvent && !timeoutEvent` case fall through, but it appears that that can never happen.

ktlim (commented):
Minor typo.

Please squash the first two commits together.

bin/telemetry_sim_ATS_1VisitNight
menanteau: That's right. This is a stand-alone simulation.

python/HeaderService/hslib.py
menanteau: fixed

python/HeaderService/hslib.py
menanteau: It should never happen, but I added and logger warning anyway.

PR #14, opened 2019-04-03
DM-17937: Clean ups

menanteau:


setup.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Build the structure for etc folder
+etc_dirs = ['etc/ATSCam','etc/TestCamera/E2V','etc/TestCamera/ITL','etc/conf']
+data_files = [("",["setpath.sh"])]
+for edir in etc_dirs:
+    data_files.append((edir,glob.glob("{}/*".format(edir))))
+
+# The main call
+setup(name='HeaderService',
+      version ='OCS_IntSpecICC',
```
mareuter: Is this really tracking the XML version? While I know that this package relies more on specific XML, how does the version change if changes are only code but not the XML?

mareuter:
General comment: Were the deleted _bin_ files moved to another package or deleted permanently?

setup.py
menanteau: Same as with the salpytool version. This is a leftover from a previous integration activity. Unfortunately we do not have a method/definition to capture versions. Open to suggestions.

timj: The normal LSST way is to define the version in terms of the git SHA1. @ktlim may have opinions on this but for HEADVER it might be easier if the version was a number that you manually increased every single time you made a commit to something that would affect the header itself. You could even adopt semantic versioning so that I could understand that going from v1 to v2 might require significant code logic changes in my metadata translator whereas 1.1.5 to 1.1.6 probably means a tweak to a comment field.  I think that would be significantly easier for me to understand in the parser than a git SHA1. We could use git tags as well that match the semantic version but I would really want the semantic version in HEADVER to not be the same as the semantic version for header service itself.

python/HeaderService/hutils.py
```diff
@@ -213,13 +213,23 @@ def get_image_size_from_imageReadoutParameters(myData):
...
-    LOGGER.info("Will start web server on dir: {}".format(dirname))
-    subprocess.Popen([exe, dirname, str(port_number)])
-    time.sleep(1)
-    LOGGER.info("Done Starting web server")
-
+    import sys
+    # Get the system's python
+    python_exe = sys.executable
+    # Make sure there isn't another process running
+    cmd = "ps -ax | grep {0} | grep -v grep | awk '{{print $1}}'".format(httpserver)
```
timj, line 222: I think this code would be significantly more robust if you used the Python `psutil` package.

setup.py
timj: and of course HEADVER for LATISS will be different to HEADVER for lsstCam.

python/HeaderService/hutils.py
menanteau: The plan is to move away from the HS starting the www server so we should stop doing this soon. In addtion psutil is not yet installed on the ATHS machine.

mareuter (approved):
Looks fine to me now with the versions changed.

PR #19, opened 2019-04-12
Tickets/dm 19127

menanteau:


bin/LFO_ATHeaderService_EventLogger
```diff
@@ -1,22 +1,30 @@
...
+import SALPY_ATHeaderService
 import HeaderService.hutils as hutils
+import datetime
 
-LOGGER = hutils.create_logger(level=logging.NOTSET, name='HEADERSERVICE')
+LOGGER = hutils.create_logger(level=logging.NOTSET, name='ATHEADERSERVICE')
+logname = "ATHS_LFO_{}.log".format(datetime.datetime.now().strftime("%Y-%m-%d-%H%M"))
+
+formatter = logging.Formatter("[%(asctime)s] [%(levelname)s] %(message)s")
+fh = logging.FileHandler(logname)
```
srp3rd, line 13: If you want to have the date there, and I suggest using TimedRotatingFileHandler instead;  it'll change the name of the log and append the date at intervals you want, AFTER the new file is created. So, for example, if you have the log name "foo", you can set it up so that at midnight, it'll change the name of the log file to "foo_2019-04-11", and start a new log named "foo" for the current day.   I don't know how much of a backlog of files it keeps as default.

If you do it using the method above, it marks it as the date the service created the log file, and it may be many months between restarts, making it a bit more difficult to decide which log file has the information you want in it.

srp3rd (approved):
Take a look at the pending comment, but the rest looks fine.

PR #20, opened 2019-04-12
Updating EXPTIME, DATE-BEG and its derived values

menanteau:


etc/ATSCam/primary_hdu.header
```diff
@@ -7,9 +7,12 @@ ORIGIN  = 'LSST DM Header Service'/ FITS file originator
          ---- Date, night and basic image information ----
 DATE    =                      / Creation Date and Time of File
 DATE-OBS=                      / Date of the observation (image acquisition)
+DATE-BEG=                      / Time at the start of integration
 DATE-END=                      / end date of the observation
 MJD     =                      / Modified Julian Date that the file was written
 MJD-OBS =                      / Modified Julian Date of observation
+MJD-BEG =                      / Modified Julian Day derived from DATE-BEG
```
timj: I'd say "corresponding to" rather than "derived from" but I'm not hung up on that. More importantly it's "Date" not "day" (day implies integer).

etc/ATSCam/primary_hdu.header
```diff
@@ -78,6 +81,7 @@ LINSTAGE=                      / Linear Stage
 
         ---- Exposure-related information ----
 EXPTIME =                      / Exposure Time in Seconds
+SHUTTIME=                      / Shutter exposure Time in Seconds
```
timj: I don't think "Time" needs to be capitalized.

python/HeaderService/hslib.py
```diff
@@ -343,11 +343,15 @@ def collect(self, keys):
     def collect_from_HeaderService(self):
         """Collect and update custom meta-data generated or transformed by the HeaderService"""
         self.DATE_OBS = hutils.get_date_utc(self.metadata['DATE-OBS'])
+        self.DATE_BEG = hutils.get_date_utc(self.metadata['DATE-BEG'])
         self.DATE_END = hutils.get_date_utc(self.metadata['DATE-END'])
         self.metadata['DATE-OBS'] = self.DATE_OBS.isot
+        self.metadata['DATE-BEG'] = self.DATE_BEG.isot
         self.metadata['DATE-END'] = self.DATE_END.isot
         self.metadata['DAYOBS'] = self.DATE_OBS.datetime.strftime('%Y%m%d')
```
timj: DAYOBS is not derived by the header service. It comes from the camera.  Maybe I'm misunderstanding this file.

python/HeaderService/hslib.py
```diff
@@ -343,11 +343,15 @@ def collect(self, keys):
     def collect_from_HeaderService(self):
         """Collect and update custom meta-data generated or transformed by the HeaderService"""
         self.DATE_OBS = hutils.get_date_utc(self.metadata['DATE-OBS'])
+        self.DATE_BEG = hutils.get_date_utc(self.metadata['DATE-BEG'])
```
timj, line 346: I'm pretty sure that @ktlim wants us to use TAI in all our headers (so TIMESYS="TAI").

timj (approved):
I have a couple of comments and a question and my question may entirely be down to me misunderstanding something.

etc/ATSCam/primary_hdu.header
menanteau: Changed Day --> Date

etc/ATSCam/primary_hdu.header
menanteau: fixed

python/HeaderService/hslib.py
menanteau: DAYOBS it is not coming from camera (at least at the moment) it is derived the startIntegration timeStamp that generates DATE-OBS.

python/HeaderService/hslib.py
menanteau: That's the plan, but we are not there yet.

python/HeaderService/hslib.py
timj: That can't be right. Camera has to publish all the components that form part of OBSID. Where do you get SEQNUM from? It's critical for the self-consistency of headers that all the components of OBSID and OBSID itself come from the same source of truth.  Note also that DAYOBS is not simply YYYMMDD from DATE-OBS because there is a UTC offset to account for the Chilean timezone.

menanteau: Camera is not publishing this yet, therefore the UGLY hack. This needs to be fix soon, once camera, which is the source of truth, publishes this too. See here for the ugly hack: https://github.com/lsst-dm/HeaderService/blob/tickets/DM-19258/python/HeaderService/hslib.py#L358

timj: Please fix DAYOBS in the same hacky way you fix SEQNUM.  You have no way of knowing the UTC offset that camera are using so if you are hacking OBSID up to get seqnum it's not a hardship to get the correct DAYOBS in the same way.

menanteau: OK -- done. Please remember that we need to fix this hack in the future.

PR #23, opened 2019-05-31
DM-19886: Remove special case for blank comment keyword

menanteau:


python/HeaderService/hutils.py
```diff
@@ -97,15 +97,14 @@ def read_head_template(fname, header=None):
...
-    lines = [l.rstrip() for l in lines if l[0:3] != 'END' and l[0:8] != ' '*8]
+    lines = [l.rstrip() for l in lines if l[0:3] != 'END']
 
     # if header is None an empty FITSHDR is created
     hdr = fitsio.FITSHDR(header)
 
     for l in lines:
         hdr.add_record(l)
-
+
```
timj: This extra space should not be included. Are you running flake8?

timj (approved):
You might want to squash the two commits (and I'm not sure if the change to GROUPID is relevant here). There is a flake8 anomaly that needs fixing.

python/HeaderService/hutils.py
menanteau: Fixed.

PR #24, opened 2019-06-06
DM-19882: YAML format for metadata files

menanteau:


timj (commented):
I'll review this tomorrow, but please can you add a travis check for flake8 compliance (as we do in other packages) as the code does not match the DM standard.

etc/dev/curated_example.yml
```diff
@@ -0,0 +1,1007 @@
+# --- Common Meta-data for all PRIMARY HDUs
+PRIMARY_COMMON:
+ SIMPLE:
```
timj: As I said during our meeting, using dict keys here does not round trip properly. If I load this YAML file the values for `COMMENT` and `None` get overwritten such that you only get the final definitions. The order is retained because python >= 3.6 now gives you ordered dicts but that still does not match our goals. The header cards need to be stored in `PRIMARY_COMMON` as elements of a list where each element defines the header item.

Something like:

```yaml
PRIMARY_COMMON:
  - keyword: SIMPLE
    value: True
    comment: "A comment"
  - keyword: BITPIX
    value: 16
    comment: "Another comment"
```

etc/dev/make_yaml_example.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+import fitsio
+
+# Define header using an example created by the HeaderService
+header_file = "AT_O_20190531_000001.header"
+header_primary = fitsio.read_header(header_file, ext='PRIMARY')
+
+header = header_primary
+print ("{}:".format(header["EXTNAME"]))
```
timj: PEP8 problem here with the space. Consider switching to f-strings.

etc/dev/make_yaml_example.py
```diff
@@ -0,0 +1,49 @@
...
+    print ("R01S01_{}:".format(header['EXTNAME']))
+
+    # Get all records
+    recs = header.records()
+    for rec in recs:
+        name = rec['name']
+        value = rec['value']
+        comment = rec['comment']
+        print (" {}:".format(rec['name']))
+        if type(value) is str:
```
timj: DM code seems to prefer

```python
if isinstance(value, str):
    ...
```

etc/dev/make_yaml_example.py
```diff
@@ -0,0 +1,49 @@
...
+            print ("  value: \"{}\"".format(rec['value']))
```
timj: `value` has already been retrieved from `rec` so it would be clearer if `value` was used here as well.

timj (changes requested):
@ktlim will probably also want to comment but I have two things that need changing:

* The YAML as currently uses a dict-based scheme that results in the comments (blank and COMMENT) being lost.
* The YAML serialization of the header should be done by first constructing a python data structure, and then doing `yaml.dump()`.  We should not be constructing the YAML strings ourselves.  Doing it via `yaml.dump` guarantees that our output is valid YAML and also guarantees that what you are writing out is what you are going to be reading in.

Finally, I will reiterate that the code in this package should conform to DM coding standards and this should be ensured by adding travis flake8 checks to this repository.

etc/dev/curated_example.yml
menanteau: I've implemented the code to write the meta-data as a list structure instead of a dictionary. For now I'm letting yaml (via dump and load) to cast the values to the proper type.  I'd assume that this is the desired behavior. 
Therefore: 
- When a keyword in undef, such as the case of the blank comment  we get `null`
- For `COMMENTS` the keyword  is actually `COMMENTS`
- When a value is not defined, we get `value=null`
Now we can have multiple `COMMENTS` and blank lines. Order of the header is preserved too.

timj (approved):
Much better. Thank you.

timj:
PS you will need to rebase and resolve the conflict.

PR #26, opened 2019-06-13
Adding container file for the HeaderService

menanteau:


Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+
+# Definitions
+ARG FITSIO_VERSION=1.0.4
+ARG SAL_VERSION=3.10.0_001
+ARG OSPL_VERSION=6.9.0
+ARG GIT_LSST="https://github.com/lsst-dm"
+ARG INSTALL_PATH=/opt/lsst
+ARG HEADERSERVICE_VERSION=0.9.7
+ARG SALPYTOOLS_VERSION=0.9.7
+ARG HSUSER=felipe
```
srp3rd: This should be a "standard" user;  we're using "iip" for all the services in ctrl_iip.   Doesn't have to be "iip", but probably shouldn't be your own either.

Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+RUN pip3 install pyyaml
+RUN pip3 install fitsio==$FITSIO_VERSION
+
+# Make python3 link
+RUN ln -s /usr/bin/python3.6 /usr/bin/python3
+
+# ---------------------
+# ts_sal rpms
+# Add the lsst-ts repo
+RUN wget https://lsst-web.ncsa.illinois.edu/~felipe/packages/lsst-ts.repo
```
srp3rd: This should be retrieved from another location, other than your own home directory. I suggest somewhere on GitHub, or in the lsst nexus repo.

Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+    ATSpectrograph-$SAL_VERSION \
+    ATTCS-$SAL_VERSION \
+    # Needed by ATArchiver
+    CatchupArchiver-$SAL_VERSION \
+    MTArchiver-$SAL_VERSION \
+    PromptProcessing-$SAL_VERSION
+
+
+# Get the setup conf
+RUN wget https://lsst-web.ncsa.illinois.edu/~felipe/packages/setup_SAL.env -O /opt/lsst/setup_SAL.env
```
srp3rd: Same comment as above.

Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+RUN chown $HSUSER:$HSUSER /opt
+
+ENV USER $HSUSER
+ENV HOME /home/$HSUSER
+ENV SHELL /bin/bash
+
+USER $HSUSER
+WORKDIR /home/$HSUSER
+
+
```
srp3rd, line 136: Is it intended that the HeaderService be running when this container is invoked?  If so, you'll need to add that.  I built the container and ran it, and didn't see the HeaderService running.   If not, make a note of how to start the HeaderService after the container is started, so people know how to do that.

Dockerfile
menanteau: Thanks, I changed the default to something more formal, like `headerservice`

Dockerfile
menanteau: Moved them to https://github.com/menanteau/lsst_ts_install_patch

Dockerfile
menanteau: No, that's not the intention yet, so I'm adding a note to the docker file.

PR #27, opened 2019-06-13
Update to utility scripts to compare FITS files and assemble them

menanteau:
Update utility scripts to compare FITS files and assemble them from scratch


bin/assemble_FITS
```diff
@@ -17,15 +17,15 @@ compress_choices = [None, 0, 11, 'RICE_1', 'PLIO_1', 21, None,
...
+    # 1. We make a proto-parse use to read in the default yaml configuration
+    # file, Turn off help, so we print all options in response to -h
     conf_parser = argparse.ArgumentParser(add_help=False)
     conf_parser.add_argument("-c", "--config", help="assemble FITS config file")
     args, remaining_argv = conf_parser.parse_known_args()
 
     # If we have -c or --config, then we proceed to read it
     if args.config:
-        conf_defaults = yaml.load(open(args.config))
+        conf_defaults = yaml.safe_load(open(args.config))
```
timj: Doesn't this leak a file descriptor?

bin/write_segment_2FITS
```diff
@@ -18,7 +18,7 @@ out_name_fz = "{}.fits.fz".format(fname)
 fd = open(fname, 'rb')
```
timj: Use a context manager here.

timj:
A general comment, but if you are writing Python executable scripts that do not have `.py` extensions, you will need to update your travis file to run flake8 on each one explicitly -- otherwise they will be ignored from the checks (and the files aren't compliant from a quick glance).

bin/write_segment_2FITS
menanteau: Changed to a context manager 
```with open(fname, 'rb') as fb:```

File: bin/assemble_FITS (hunk 1 above)
Comment (menanteau) [reply bin/assemble_FITS]: Changed to: 

```
with open(args.config) as conf:
          conf_defaults = yaml.safe_load(conf)
```

PR #29, opened 2019-06-14
Move imports to the top

menanteau:


python/HeaderService/hutils.py
```diff
@@ -192,13 +194,11 @@ def get_date_utc(timeStamp=None, format='isot'):
...
 
-    from astropy.time import Time
-    from datetime import datetime
     if timeStamp is None:
-        utc_time = (datetime.utcnow()).isoformat()
+        utc_time = (datetime.datetime.utcnow()).isoformat()
     else:
-        utc_time = datetime.utcfromtimestamp(timeStamp).isoformat()
-    t = Time(utc_time, format=format, scale='utc')
+        utc_time = datetime.datetime.utcfromtimestamp(timeStamp).isoformat()
```
timj: You can avoid the unnecessary stringification here by going directly to Astropy Time:
```
t = astropy.time.Time(timeStamp, format="unix")
```

python/HeaderService/hutils.py
```diff
@@ -192,13 +194,11 @@ def get_date_utc(timeStamp=None, format='isot'):
 
     """
 
-    from astropy.time import Time
-    from datetime import datetime
     if timeStamp is None:
-        utc_time = (datetime.utcnow()).isoformat()
+        utc_time = (datetime.datetime.utcnow()).isoformat()
```
timj: Avoid the extra stringification by doing:
```
t = astropy.time.Time.now()
```

python/HeaderService/hutils.py
```diff
@@ -140,7 +143,6 @@ def md5Checksum(filePath, blocksize=1024*512):
 
 
 def get_obsnite(date=None, thresh_hour=14, format='{year}{month:02d}{day:02d}'):
```
timj, line 159: Is this function used any more now that DAYOBS is being extracted from OBSID?

python/HeaderService/hutils.py
```diff
@@ -120,7 +124,6 @@ def get_values(header):
 
 
 def elapsed_time(t1, verb=False):
```
timj, line 126: Methods need doc strings.

python/HeaderService/hutils.py
```diff
@@ -33,6 +33,10 @@
 import itertools
 import copy
 from .camera_coords import CCDGeom
+import astropy
```
timj: The usual convention in Astropy is to do `from astropy.time import Time`

timj (approved):
Looks ok. I think the astropy time code should be cleaned up to stop generating an intermediary ISO format string that we don't need.

python/HeaderService/hutils.py
menanteau: Thanks for the suggestion, it's been implemented.

python/HeaderService/hutils.py
menanteau: Done too.

python/HeaderService/hutils.py
menanteau: Not anymore, your comment implies that it is okay to get rid of it. Therefore that what I'll do.

python/HeaderService/hutils.py
menanteau: Updated to `from astropy.time import Time` at the top of the file.

python/HeaderService/hutils.py
menanteau: Added doc string.

PR #30, opened 2019-06-19
Tickers/dm 20134

menanteau:


python/HeaderService/hscalc.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    if timeStamp is None:
+        t = Time.now()
+    else:
+        t = Time(timeStamp, format="unix")
+    return t
+
+
+def get_radec_from_altaz(alt, az, obstime, lat=-30.244639, lon=-70.749417, height=2663.0):
```
patrickingraham, line 63: Where are you storing the real values for lat/long ? They should be stored elsewhere and always passed rather than rely on defaults.

python/HeaderService/hscalc.py
```diff
@@ -0,0 +1,105 @@
...
+    ----------
+
+    alt : float
+        The Altitude (angle) for the object
+    az : float
+        The Azimuth (angle) for the object
+    obstime: float
+        The time of the observation
+    lat: float
+        Optional, the geografic latitude in degrees
```
patrickingraham: "geographic"

python/HeaderService/hscalc.py
```diff
@@ -0,0 +1,105 @@
...
+    lon: float
+        Optional, the geografic longitude in degrees
+    height: float
+        Optional, the height in meters
+
+    Returns
+    -------
+
+    ra: float
+       The Right Ascencion
```
patrickingraham: ascension

python/HeaderService/hslib.py
```diff
@@ -359,23 +360,49 @@ def collect_from_HeaderService(self):
...
         self.metadata['FILENAME'] = self.filename_FITS
         # THIS IS AN UGLY HACK TO MAKE IT WORK SEQNUM
         # FIX THIS -- FELIPE, TIAGO, MICHAEL and TIM J. are accomplices
         self.metadata['SEQNUM'] = int(self.metadata['OBSID'].split('_')[-1])
         self.metadata['DAYOBS'] = self.metadata['OBSID'].split('_')[2]
 
+        # The EL/AZ at start
+        if 'ELSTART' and 'AZSTART' in self.metadata:
+            LOGGER.info("Computing RA/DEC from ELSTART/AZSTART")
+            ra, dec = hscalc.get_radec_from_altaz(alt=self.metadata['ELSTART'],
```
patrickingraham, line 387: this is relying on your default position in the function rather than the top-level value that should be passed. I know it's annoying but defaults are asking for trouble later on..

python/HeaderService/hslib.py
```diff
@@ -359,23 +360,49 @@ def collect_from_HeaderService(self):
...
+                                                  obstime=self.DATE_BEG)
+            self.metadata['RASTART'] = ra
+            self.metadata['DECSTART'] = dec
+        else:
+            LOGGER.info("No 'ELSTART' and 'AZSTART'")
+
+        # The EL/AZ at end
+        if 'ELEND' and 'AZEND' in self.metadata:
+            LOGGER.info("Computing RA/DEC from ELEND/AZEND")
+            ra, dec = hscalc.get_radec_from_altaz(alt=self.metadata['ELEND'],
```
patrickingraham, line 401: same as above regarding defaults

etc/ATSCam/primary_hdu.header
```diff
@@ -30,13 +30,14 @@ OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
...
         ---- Pointing info, etc. ----
-RATEL   =                      / Telescope RA of observation
-DECTEL  =                      / Telescope DEC of observation
-ROTPATEL=                      / Telescope Rotation
-ROTCOORD= 'sky'                / Telescope Rotation Coordinates
-RA      =                      / RA of Target
-DEC     =                      / DEC of Target
+RA      =                      / Telescope RA 
+DEC     =                      / Telescope DEC
+RASTART =                      / RA of Target at start
```
patrickingraham: It is important to say these are calculated from the telelescope Az/El position!

etc/ATSCam/primary_hdu.header
```diff
@@ -30,13 +30,14 @@ OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
 OBSGEO-Z=          -3195172.08 / [m] Z-axis Geocentric coordinate
 
         ---- Pointing info, etc. ----
-RATEL   =                      / Telescope RA of observation
-DECTEL  =                      / Telescope DEC of observation
-ROTPATEL=                      / Telescope Rotation
-ROTCOORD= 'sky'                / Telescope Rotation Coordinates
-RA      =                      / RA of Target
-DEC     =                      / DEC of Target
+RA      =                      / Telescope RA
```
patrickingraham: Need to say this comes from the pointing component

etc/conf/atTelemetry.yaml
```diff
@@ -142,33 +143,13 @@ telemetry:
...
-  collect_after_event: start_collection_event
-
- ROTPATEL:
-  device: ATPtg
-  Stype: Event
-  topic: currentTarget
-  value: rotPA
-  collect_after_event: start_collection_event
+# --- Will come from ATMCS rotator angle when available
+# ROTPA:
```
patrickingraham, line 147: There is also the rotator start/end values that come from the ATMCS

patrickingraham (changes requested):
This is just a quick once over.

python/HeaderService/hscalc.py
timj: They should definitely be guaranteed to match the OBSGEO values in the header. We can't afford to get into odd inconsistencies where the header has one location and the derived values use different values.  In an ideal world the pointing task would be using the same telescope location as everyone else so the place the pointing task gets the location should be the same place the header service gets it from (and if we ever change that location it should be part of configuration control and changed in a single place).

python/HeaderService/hscalc.py
menanteau: Thanks for catching this, fixed.

python/HeaderService/hscalc.py
menanteau: Fixed too.

python/HeaderService/hslib.py
menanteau: Now the top value, as defined by the header template is passed instead.

python/HeaderService/hslib.py
menanteau: Fixed as above.

etc/ATSCam/primary_hdu.header
menanteau: Comment added to the header template.

etc/ATSCam/primary_hdu.header
menanteau: Comment added to the header template.

etc/conf/atTelemetry.yaml
menanteau: Yep, but we should add that to another ticket.

python/HeaderService/hscalc.py
menanteau: Now it is passing the values defined on the primary header template.

bin/telemetry_sim_ATS_1VisitNight
```diff
@@ -214,8 +214,8 @@ if __name__ == "__main__":
 
         # end telemetry for EL and AZ from the ATMCS.mountEncoders
         kw = {'timestamp': time.time(),
-              'elevationCalculatedAngle': args.el*1.2,
-              'azimuthCalculatedAngle': args.az*1.2}
+              'elevationCalculatedAngle': args.el*1.001,
```
patrickingraham, line 222: I realize you hardcoded this because you don't have the proper telemetry to put it in, however, you should just add a note saying so. People often use something like, "TODO: Update to use end value once proper telemetry is available"

etc/ATSCam/primary_hdu.header
```diff
@@ -30,12 +30,12 @@ OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
...
         ---- Pointing info, etc. ----
-RA      =                      / Telescope RA 
-DEC     =                      / Telescope DEC
-RASTART =                      / RA of Target at start
-DECSTART=                      / DEC of Target at start
-RAEND   =                      / RA of Target at end
-DECEND  =                      / DEC of Target at end
+RA      =                      / Telescope RA from pointing component
+DEC     =                      / Telescope DEC from pointing component
+RASTART =                      / RA from telelescope AZ/EL position at start
```
patrickingraham: "RA of telescope calculated from AZSTART and ELSTART"

etc/ATSCam/primary_hdu.header
```diff
@@ -30,12 +30,12 @@ OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
 OBSGEO-Z=          -3195172.08 / [m] Z-axis Geocentric coordinate
 
         ---- Pointing info, etc. ----
-RA      =                      / Telescope RA 
-DEC     =                      / Telescope DEC
-RASTART =                      / RA of Target at start
-DECSTART=                      / DEC of Target at start
-RAEND   =                      / RA of Target at end
-DECEND  =                      / DEC of Target at end
+RA      =                      / Telescope RA from pointing component
```
patrickingraham: RA commanded from pointing component
--> It's important to know it's the commanded position, not necessarily the actual position. No need to say "telescope"

etc/ATSCam/primary_hdu.header
```diff
@@ -30,12 +30,12 @@ OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
...
+DEC     =                      / Telescope DEC from pointing component
```
patrickingraham: DEC commanded from pointing component

etc/ATSCam/primary_hdu.header
```diff
@@ -30,12 +30,12 @@ OBSGEO-Y=          -5208470.95 / [m] Y-axis Geocentric coordinate
...
+RASTART =                      / RA from telelescope AZ/EL position at start
+DECSTART=                      / DEC from telelescope AZ/EL position at start
+RAEND   =                      / RA from telelescope AZ/EL position at end
+DECEND  =                      / DEC from telelescope AZ/EL position at end
```
patrickingraham: see comments above and apply here.

patrickingraham (commented):
Looks much better. Just a few minor documentation and description changes to make then it's good to go.

bin/telemetry_sim_ATS_1VisitNight
menanteau: This simple script `telemetry_sim_ATS_1VisitNight` just makes up telemetry to test the creation and receiving of messages through SAL, so proper telemetry is not expected to ever happen in this piece of code. The `1.001` factor is in place to change the values between the `START` and `END` and make sure that the computed `RASTART/RAEND` and `DECSTART/DECEND` are actually different during testing. I will add WARNING to comments to clarify what's happening here.

etc/ATSCam/primary_hdu.header
menanteau: Changed to `"RA of telescope from AZSTART and ELSTART"` in order to fit the 80 char limit for the FITS card.

etc/ATSCam/primary_hdu.header
menanteau: Changes to `"RA commanded from pointing component"` and `"DEC commanded from pointing component"`

etc/ATSCam/primary_hdu.header
menanteau: Changed.

etc/ATSCam/primary_hdu.header
menanteau: Done.

PR #31, opened 2019-07-09
Changes for compatibility with ts_xml 3.10.2

menanteau:


etc/conf/atTelemetry.yaml
```diff
@@ -58,21 +58,21 @@ telemetry:
...
-  value: timeStamp
+  value: timeStampAcquisitionStart
   collect_after_event: start_collection_event
 
  DATE-END:
   device: ATCamera
   Stype: Event
   topic: endReadout
-  value: timeStamp
+  value: timeStampAcquisitionStart
```
patrickingraham, line 75: shouldn't this be at the end of the acquisition rather than the start?

etc/ATSCam/primary_hdu.header
```diff
@@ -48,7 +48,7 @@ AZEND   =                      / [deg] Telescope azimuth angle at end
 AMEND   =                      / Airmass at end
 
         ---- Image-identifying used to build OBS-ID ----
-TELCODE = 'AT'                 / The code for the telecope
+TELCODE =                      / The code for the telecope (AT/CC/MC)
 CONTRLLR=                      / The controller (e.g. O for OCS, C for CCS)
```
patrickingraham: ACCS/CCS

The ACCS is used to control the ATCAM.

etc/conf/atTelemetry.yaml
```diff
@@ -10,7 +10,7 @@ logfile: 'www/HeaderService.log'
 vendor: ITL
 
 # The ts_xml version we are using -- for now this is just curated and need to be changed manually when we change version
-ts_xml: "3.10.0"
+ts_xml: "3.10.2"
 
 # port_number for webserver
 port_number: 8000
```
patrickingraham, line 16: This probably shouldn't be hardcoded

bin/telemetry_sim_ATS_1VisitNight
```diff
@@ -180,8 +180,11 @@ if __name__ == "__main__":
         LOGGER.info("Sending starIntegration")
         kw = {'imageSequenceName': args.imageSequenceName,
               'imageName': imageName,
-              'imageIndex': k+1,
-              'timeStamp': timeStamp,
+              'imageNumber': k+1,
```
patrickingraham, line 183: Sorry, I'm just not quite understanding this.
what is k? and why it is k+1 ?

patrickingraham (changes requested):
Might not require changes but please answer my questions!

bin/telemetry_sim_ATS_1VisitNight
menanteau: When requesting more than one exposure per sequence.

etc/conf/atTelemetry.yaml
menanteau: Yes, but we don't have any other way to get it at the moment.

etc/ATSCam/primary_hdu.header
menanteau: Is this comment for the CONTRLLR key on the template?

etc/conf/atTelemetry.yaml
menanteau: Yes, we discussed this yesterday at the CAP meeting. We have been using the wrong timeStamp to register the END event for the observing window all this time. The recent change in the name of the variable by Tony has made this clear for us now. We have requested Tony to add a timeStamp for the endofReadout event.

etc/ATSCam/primary_hdu.header
patrickingraham: Yes. 
"The controller (e.g. O for OCS, C for (A)CCS)

menanteau: Thanks for the clarification. Fixed.

PR #32, opened 2019-07-24
Tickets/dm 20479

menanteau:


Dockerfile
```diff
@@ -1,136 +1,73 @@
...
-# Get the setup conf
-RUN wget https://raw.githubusercontent.com/menanteau/lsst_ts_install_patch/master/setup_SAL.env -O /opt/lsst/setup_SAL.env
-# ---------------------
-
-# Add $HSUSER as user
-RUN useradd -ms /bin/bash $HSUSER
-RUN usermod -aG wheel $HSUSER
-
-# Install HeaderService
+# --- Install HeaderService and salpytools ----
```
r-owen, line 35: Are you still using salpytools? If not, please remove or correct the references in this file

bin/csc_send_to_state
```diff
@@ -0,0 +1,31 @@
...
+
+def cmdline():
+
+    import argparse
+    parser = argparse.ArgumentParser(description="Send CSC to a given State")
+
+    # The optional arguments
+    parser.add_argument("state", action="store",
+                        help="Desired State")
+    parser.add_argument("-c", "--csc_name", action="store", default='ATHeaderService',
```
r-owen: I suggest this be a required argument (changing `"-c", "--csc_name"` to `"csc_name"`) because it makes no sense to run this without specifying a CSC. Please also consider having it be the first argument, as that matches the order of the words in the command name.

bin/telemetry_sim_ATHS
```diff
@@ -0,0 +1,203 @@
...
+
+    # Camera Geometry
+    # Send sub-set of imageReadoutParameters, we populate only a few fields
+    kwRO = {'overRows': args.overv,
+            'overCols': args.overh,
+            'preCols': args.preh,
+            'readCols': _GEO.dimh,
+            'readCols2': 0,
+            'readRows': _GEO.dimv}
+    cam.evt_imageReadoutParameters.set_put(**kwRO)
```
r-owen, line 100: Consider the following variant, for clarity. It is more direct, reduces punctuation and eliminates a variable:

    cam.evt_imageReadoutParameters.set_put(overRows=args.overv,
                                           overCols=args.overh,
                                           preCols=args.preh,
                                           readCols=_GEO.dimh,
                                           readCols2=0,
                                           readRows=_GEO.dimv,
                                           force_output=True)

`force_output=True` isn't strictly necessary, but it is best practice when you want to be sure a `ControllerEvent` will be emitted. Without it, the event is only emitted if data has changed. (Note: `ControllerTelemetry.set_put` has no `force_output` argument because  telemetry topics are always output).

Similarly for the other events below....

bin/telemetry_sim_ATHS
```diff
@@ -0,0 +1,203 @@
...
+import HeaderService.hscalc as hscalc
+import HeaderService.camera_coords as camera_coords
+import random
+
+import lsst.ts.salobj as salobj
+
+LOGGER = hutils.create_logger()
+
+# Get the info from the camera_coords library for and E2V sensor
+_GEO = camera_coords.CCDGeom('E2V')
```
r-owen: Consider `CCDGEOM` instead of `_GEO`. It's a global, but it's not private (as suggested by the underscore) and the current name is ambiguous.

bin/telemetry_sim_ATHS
```diff
@@ -0,0 +1,203 @@
...
+            DAYOBS = DATE_OBS.datetime.strftime('%Y%m%d')
+        imageName = "{}_{}_{}_{:06d}".format(args.telcode, args.controller, DAYOBS, seqN)
+
+        # Send CCD temperature telemetry using a random generator
+        kwCCD = {'ccdTemp0': random.uniform(-30, 0)}
+        cam.tel_wreb.set_put(**kwCCD)
+        LOGGER.info("Sending wreb")
+
+        # The START telemetry for Airmass using ATPtg.prospectiveTargetStatus
+        kwTA = {'timestamp': time.time(),
```
r-owen: timestamps should be TAI, not UTC. I suggest:

    from lsst.ts import salobj
    timestamp=salobj.tai_from_utc(time.time())

here and every where else that you use time.time().

I plan to add a function to ts_salobj that returns the current TAI (in unix seconds) but have not yet done so.

etc/conf/atTelemetry_salobj.yaml
```diff
@@ -0,0 +1,271 @@
...
+hdu_delimiter: END
+
+# The file naming event source and definitions
+filepath: 'www/DMHS_filerepo'
+format_HDR: "{}.header"
+format_FITS: "{}.fits"
+imageName_event:
+    device: ATCamera
+    topic: startIntegration
+    value: imageName
```
r-owen: How do you handle SAL components that have non-zero index? Reading below I suspect you do not. I mention later but consider allowing the name to have an optional suffix `:index` e.g. `ATPtg:0`. Then parsing as name, integer SAL index (which defaults to 0 if unspecified).

bin/csc_send_to_state
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
r-owen, line 1: Can you please explain why you have this command-line utility? I'm guessing it is a convenience while debugging, in which case it might be better to make it a personal utility instead of part of this package. My concern is that anyone who sets up the header service will have this as an available binary and it's not a safe thing to run under normal circumstances.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import HeaderService
+import HeaderService.hutils as hutils
+import HeaderService.hscalc as hscalc
```
r-owen: I suggest you replace the above 3 lines with the following, as it reduces the chance of problems due to circular imports:

    from . import hutils
    from . import hscalc

(note that I have deleted import HeaderService entirely). This is best practice for library code, for the reason given (but such relative imports are not appropriate in bin scripts, unit tests and other code run from the command line).

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+import logging
+import os
+import sys
+import socket
+import lsst.ts.salobj as salobj
```
r-owen: I suggest this, to reduce stutter:

    from lsst.ts import salobj

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+
+    """ A Class to run and manage the Header Service"""
+
+    def __init__(self, **keys):
+
+        self.keys = keys
+        # Unpack the dictionary of **keys into variables to do:
+        # self.keyname = key['keyname']
+        for k, v in list(keys.items()):
+            setattr(self, k, v)
```
r-owen: This makes the code difficult to understand. A user will have no idea what the constructor arguments are.

It is also unsafe because an argument that has the same name as a method will result in a bug that is very difficult to track down. Such bugs could easily be introduced any time in the future by adding a new method that accidentally has the same name as a constructor argument, so it is a time bomb waiting to go off.

Best practice is to list all keyword arguments explicitly. That makes them easy to document, easy for a user to understand, provides a simple way to provide useful defaults that are self-documenting, and is the standard idiom. However, it can get repetitive if you have a *lot* of arguments and can get a bit messy if you have to validate the arguments against valid ranges, allowed choices, data types, etc.

An alternative that ameliorates these issues is to provide a json schema file that describes the arguments, including the following information for each argument:
* name
* description
* type
* restrictions such as minimums, maximums, valid choices...
* a default value, if appropriate

I have never used a schema for constructor arguments, but salobj uses it for Configurable CSCs and Scripts. This might be a case where the construction is enough like configuration to justify a schema-based approach. If you go that route then I have two suggestions:

1) Use `DefaultingValidator` in `salobj` to validate the arguments instead of the standard validator in `jsonschema`, because the latter does not set default values!

2) After you have validated the arguments as a dict, make them available in your class as as standard attributes using something like this:

        self.args = types.SimpleNamespace(**keys)

then you can get to all constructor arguments using, for example: `self.args.ra`. That is a lot nicer than `self.keys["ra"]`, almost as short as `self.ra` and is immune to name collisions now and in the future.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+import HeaderService.hscalc as hscalc
+import logging
+import os
+import sys
+import socket
+import lsst.ts.salobj as salobj
+import asyncio
+import time
+
+LOGGER = logging.getLogger(__name__)
```
r-owen: BaseCsc already includes a logger as attribute `log`. Please use it instead of creating your own.

Thus replace LOGGER.info(...) with `self.log.info(...)` and similarly in all code below. Obviously you can't have a log message until you call `super().__init__(...)` but you lose little that way.

Once you have called `super().__init__(...)` you can configure the logger just as you do now. Note that one thing you get for free using this approach is that log messages will be sent as `logMessage` events (as well as logged however else you have configured), so they show up in the EFD.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+            if c['Stype'] == 'Telemetry':
+                self.Remote_get[name] = getattr(self.Remote[devname], f"tel_{c['topic']}").get
+                LOGGER.info(f"Storing Remote.tel_{c['topic']}.get() for {name}")
+
+        # Select the start_collection channel
+        self.name_start = get_channel_name(self.start_collection_event)
+        # Select the end_collection channel
+        self.name_end = get_channel_name(self.end_collection_event)
+
+    def create_Controllers(self):
```
r-owen: `BaseCsc` already inherits from `Controller`, so `self` is already a Controller and has the necessary attributes (once you call `super().__init__(...)`. At the very least, please:

* Delete lines 235-237.
* Change all references to `self.hs_controller` to `self`

In addition I suggest you ditch this method and move lines 240-241 to the body of your `__init__` method since there's almost nothing left for this method to do.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+        # Make sure the directory exists
+        dirname = os.path.dirname(self.logfile)
+        if dirname != '':
+            self.check_outdir(dirname)
+        hutils.create_logger(logfile=self.logfile, level=self.loglevel,
+                             log_format=self.log_format, log_format_date=self.log_format_date)
+        LOGGER.info("Logging Started")
+        LOGGER.info(f"Will send logging to: {self.logfile}")
+
+    def create_BaseCsc(self):
```
r-owen, line 186: Please consider moving these few lines of code into `__init__`. Reasons:
* It makes the API harder to use incorrectly. This method should only be called once, during construction, so it should not be a public method.
* It makes the code easier to understand and more idiomatic by moving the call to `super().__init__(...)` into the body of the `__init__` method.
* It simplifies the code by eliminating a method.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+    def get_channels(self):
+        """Extract the unique channels by topic/device"""
+        LOGGER.info("Extracting Telemetry channels from telemetry dictionary")
+        self.channels = extract_telemetry_channels(self.telemetry,
+                                                   start_collection_event=self.start_collection_event,
+                                                   end_collection_event=self.end_collection_event,
+                                                   imageParam_event=self.imageParam_event)
+        # Separate the keys to collect at the 'end' from the ones at 'start'
+        t = self.telemetry  # Short-cut
+        self.keywords_start = [k for k in t.keys() if t[k]['collect_after_event'] == 'start_collection_event']
```
r-owen: Consider the following to avoid the need for an explicit lookup (and also reduce the need for a shortcut):

    self.keywords_start = [key for key, value in self.telemetry.items() if value["collect_after_event"] == "start_collection_event"]

If you are keen to do the lookup yourself, then it is more idiomatic to take advantage of the fact that `dict` acts as an iterator over its keys: `[k for k in t if ...]` instead of `[for k in t.keys() if ...]`.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+        self.keywords_end = [k for k in t.keys() if t[k]['collect_after_event'] == 'end_collection_event']
```
r-owen: Are you sanity-checking the values in ['collect_after_event'] and failing if they aren't one of the allowed choices? I didn't see the code for that. If it's missing, please add it. (Or if only 2 values are allowed, consider using a boolean instead of a string).

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+
+    def check_outdir(self, filepath):
+        """ Make sure that we have a place to put the files"""
+        if not os.path.exists(filepath):
+            os.makedirs(filepath)
+            LOGGER.info("Created dirname:{}".format(filepath))
```
r-owen: Consider an f string for clarity and brevity: `LOGGER.info(f"Created dirname: {filepath}")` or, I hope, `self.log.info(...)`

and similarly for other log messages.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        LOGGER.info("*** Starting Connections for Meta-data ***")
+        # The list containing the unique devices (CSCs) to make connection
+        self.devices = []
+        # The dict containing all of the threads and connections
+        self.Remote = {}
+        self.Remote_get = {}
+        for name, c in self.channels.items():
+            devname = c['device']
+            # Make sure we only crete these once
```
r-owen: crete -> create

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+            if devname not in self.devices:
+                self.devices.append(devname)
+                self.Remote[devname] = salobj.Remote(domain=self.domain, name=devname, index=0)
```
r-owen, line 215: This code assumes that the SAL components you listen to will always use index=0. I worry that you may find this too restrictive in the future. Consider allowing a device name of the form `name:index ` where `:index` is optional. Use `re` or `split` to parse the result. We have plenty of existing code that uses this convention including the `SetSummaryState` script in `ts_standardscripts` and you can borrow parsing code from that.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        Collect and update custom meta-data generated or transformed by
+        the HeaderService
+        """
+
+        # Reformat and calculate dates based on different timeStamps
+        self.DATE = hscalc.get_date_utc(self.metadata['DATE'])
+        self.DATE_OBS = hscalc.get_date_utc(self.metadata['DATE-OBS'])
+        self.DATE_BEG = hscalc.get_date_utc(self.metadata['DATE-BEG'])
+        self.DATE_END = hscalc.get_date_utc(self.metadata['DATE-END'])
```
r-owen: I am pretty sure LSST chose to use TAI for all timestamps. In which case these should be TAI, not UTC and you should set FITS header `TIMESYS=TAI`.

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,500 @@
...
+import sys
+import socket
+import lsst.ts.salobj as salobj
+import asyncio
+import time
+
+LOGGER = logging.getLogger(__name__)
+
+
+class HSWorker(salobj.BaseCsc):
```
r-owen, line 33: In this package I see three classes called HSWorker (though this seems to be the main one, and the others are simple examples). Please consider tweaking the names of the classes in the example files to avoid confusion.

r-owen (changes requested):
Some changes requested, including:
- In HSWorker don't create a Controller that duplicates the BaseCsc.
- Use the logger provided by BaseCsc instead of creating your own.
- Use explicit keywords instead of **keys to construct HSWorker (or at least use a schema). I think this is the only request that would not be easy to implement.

Dockerfile
menanteau: Yes, still using salpytools for now. I would like to drop it the near future.

bin/csc_send_to_state
menanteau: Sure, I've changed this as you suggested

bin/telemetry_sim_ATHS
menanteau: I actually prefer the current variant, and I find it easier to read. I added the `force_output=True` as suggested.

bin/telemetry_sim_ATHS
menanteau: Thanks, let me know when that function is ready to use it. I've modified the code to use `salobj.tai_from_utc(time.time())` instead.

bin/csc_send_to_state
menanteau: Beyond my own debugging, at the moment it is required that each CSC transition itself through the different states.

python/HeaderService/hslib_salobj.py
menanteau: Thanks, I've updated the code. However I think I still need the:
```
import HeaderService
```
as this is used down bellow as:
``` # Load up the header templates
        self.HDR = HeaderService.HDRTEMPL_ATSCam(vendor=self.vendor,
                                                 write_mode=self.write_mode,
                                                 hdu_delimiter=self.hdu_delimiter)
```

Do you have any suggestions?

python/HeaderService/hslib_salobj.py
menanteau: Thanks for the suggestion, this part has been bothering me for some time. I'll talk or slack you about the best procedure to implement.

python/HeaderService/hslib_salobj.py
menanteau: Done.

python/HeaderService/hslib_salobj.py
menanteau: Done

python/HeaderService/hslib_salobj.py
menanteau: Thinking of a vacation year ago :)

python/HeaderService/hslib_salobj.py
menanteau: Thanks for the suggestion, used the first option.

python/HeaderService/hslib_salobj.py
menanteau: As suggested by Tim J, we are using UTC times for the FITS headers at the moment.

timj: LSE-400 says we should use TIMESYS=TAI so we should be switching over.

menanteau: Thanks for the clarification, will update accordingly.

python/HeaderService/hslib_salobj.py
r-owen: `HDRTEMPL_ATSCam` is defined in `hsutils` so my suggestion is replace the line below with `self.HDR = hsutils.HDRTEMPL_ATSCam(...`

At that point I think you can safely delete line 22 `import HeaderService`

python/HeaderService/hslib_salobj.py
r-owen: Should you decide to go this route at some point I suggest you use function `salobj.name_to_name_index` to split the name for you:

    name, index = salobj.name_to_name_index(devname)

menanteau: Thanks!

python/HeaderService/hslib_salobj.py
menanteau: Thanks.

python/HeaderService/hslib_salobj.py
menanteau: Done.

etc/conf/atTelemetry_salobj.yaml
```diff
@@ -0,0 +1,271 @@
...
+
+ DEC:
+  device: ATPtg
+  Stype: Event
+  topic: currentTarget
+  value: declination
+  collect_after_event: start_collection_event
+
+# --- Will come from ATMCS rotator angle when available
+# ROTPA:
```
r-owen: `rotPA` is a field of the target `currentTarget` event. Can you enable this entry?

etc/conf/atTelemetry_salobj.yaml
```diff
@@ -0,0 +1,271 @@
...
+ GROUPID:
+  device: ATCamera
+  Stype: Event
+  topic: startIntegration
+  value: imageSequenceName
+  collect_after_event: start_collection_event
+
+# ----------------------------------------
+# -- still not available from SAL --------
+# OBSTYPE:
```
r-owen: I assume making this available is work that is planned?

If there is no Jira ticket to add the feature, might you create one? If there is such a ticket then please file a ticket against HeaderService to update this code once the other ticket is implemented (and make it block against the other).

python/HeaderService/hscalc.py
```diff
@@ -49,14 +49,15 @@ def get_date_utc(timeStamp=None):
...
+       The astropy.datetime.Time object with the timeStamp requested
+       in the defined scale.
 
     """
 
     if timeStamp is None:
         t = Time.now()
     else:
-        t = Time(timeStamp, format="unix")
+        t = Time(timeStamp, format="unix", scale=scale)
```
r-owen, line 60: I am virtually certain this does not work, based on painful experience trying to get TAI as unix seconds from AstroPy. The problem is that Astropy always treats format="unix" as UTC, even if you specify a different scale. You can easily see this by creating two Times with format="unix" and different scales and observing that the difference is 0 seconds.

It does work as expected with other formats, so this is what I ended up doing in ts_ATMCSSimulator for a similar problem. Perhaps you can adapt it:
```
def curr_tai():
    """Get the current time as TAI in unix seconds.

    This is the same format used by the ATPtg CSC.
    """
    # unfortunately I can't just output curr_time.tai.unix
    # because that has the same value as curr_time.utc.unix
    curr_time = Time.now()
    tai_minus_utc = (curr_time.tai.mjd - curr_time.utc.mjd)*24*60*60
    return curr_time.utc.unix + tai_minus_utc
```

python/HeaderService/hslib_salobj.py
```diff
@@ -0,0 +1,495 @@
...
+
+    """ A Class to run and manage the Header Service"""
+
+    def __init__(self, **keys):
+
+        # Load the configurarion
+        self.conf = types.SimpleNamespace(**keys)
```
r-owen: Thank you for making this change. Please consider naming this `self.config` as when you make the header service a ConfigurableCsc that is what the argument to the configure method will be called. (And it matches are usage in other CSCs).

python/HeaderService/hutils.py
```diff
@@ -50,30 +50,49 @@
...
 
     # Set the screen handle
     sh = logging.StreamHandler(sys.stdout)
     sh.setFormatter(formatter)
+    sh.setLevel(level)
     handlers.append(sh)
-    logging.basicConfig(handlers=handlers, level=level, format=FORMAT, datefmt=FORMAT_DATE)
+    logger.addHandler(sh)
+
+    return handlers
```
r-owen: Is there any need to return the handlers? One can get them from the logger and it would simplify the code to not bother. I see you are using it below in create_logger but I suspect that is not needed. Certainly I have found that by adding a StreamHandler and setting its level I get the output I expect without calling basicConfig.

python/HeaderService/hutils.py
```diff
@@ -50,30 +50,49 @@
...
+
+
+def create_logger(logfile=None, level=logging.NOTSET, log_format=None, log_format_date=None):
+    """
+    Simple logger that uses configure_logger()
+    """
+    LOGGER = logging.getLogger(__name__)
+    handlers = configure_logger(LOGGER, logfile=logfile, level=level,
+                                log_format=log_format, log_format_date=log_format_date)
+    logging.basicConfig(handlers=handlers, level=level)
```
r-owen: Is this still needed? I was hoping that by adding a StreamHandler you would find that basicConfig was no longer needed.

r-owen (approved):
I fear you have a bug in handling TAI that I noted below and I fear you will have to deal with that. (I encourage you to verify if you got it right by making sure there is the expected difference between TAI and UTC times).

In addition I had a few small suggested changes, but they are optional. Overall this looks fine. Thank you for making the earlier changes!

etc/conf/atTelemetry_salobj.yaml
menanteau: Yes, there is a jira ticket. Waiting for Tony to implement this. Hopefully in the next xml version.

python/HeaderService/hslib_salobj.py
menanteau: Sure, will change to self.config

etc/conf/atTelemetry_salobj.yaml
menanteau: It was deemed inappropriate by Patrick and should be replaced by `ROTSTART/END` from encoders in the near future. Left as a reminder.

python/HeaderService/hscalc.py
menanteau: It seems to be working for me, for example:
```
import astropy
import time
from astropy.time import Time 

timeStamp = time.time()    
Time(timeStamp, scale='utc', format="unix").isot                                                                        
'2019-08-01T21:09:16.625'

Time(timeStamp, scale='tai', format="unix").isot   
'2019-08-01T21:09:53.625'
```

Did I miss anything?

r-owen: I guess it is expressing time in unix seconds (e.g. `.unix` instead of `.iso`) that shows the problem (the two outputs are identical in that case). If you are using the code to write ISO times then it should work fine. Thank you for checking.

python/HeaderService/hutils.py
menanteau: Thanks for noticing it. I've fixed it.

python/HeaderService/hutils.py
menanteau: Unfortunately, without the `basicConfig` call the logger doesn't show.

PR #33, opened 2019-07-30
DM-20826: Implement new header format (yaml) for AuxTel

menanteau:


python/HeaderService/hslib.py
```diff
@@ -319,7 +319,8 @@ def announce(self):
         kw = {'byteSize': bytesize,
               'checkSum': md5value,
               'generator': 'ATHeaderService',
-              'mimeType': 'FITS',
+              # Make sure this is also changed on hslib_salobj when merged
+              'mimeType': self.write_mode.upper(),
```
timj: Is this meant to be a real mime type or are you using mime type here to be an internal format name?

python/HeaderService/hutils.py
```diff
@@ -354,23 +338,31 @@ def calculate_string_header(self):
...
-            pool.map(write_header_string, args)
-            pool.close()
-            pool.join()
-        else:
-            for filename in filenames:
-                arg = filename, hstring, md5
-                write_header_string(arg)
-        return
+        # The dict where we will store the header contents
+        self.yaml_header = {}
```
timj: Does this need to be assigned to `self`? It would be less typing below to use a short local variable and then assign to `self` when the loop is finished.  This also means that if something goes wrong in the loop you don't end up with a partially populated `yaml_header`.

python/HeaderService/hutils.py
```diff
@@ -431,8 +417,8 @@ def write_header(self, filename):
         t0 = time.time()
         if self.write_mode == 'fits':
             self.write_header_fits(filename)
-        elif self.write_mode == 'string':
-            self.write_header_string(filename)
+        elif self.write_mode == 'yaml' or self.write_mode == 'yml':
```
timj: How do you get `yml` option?  Does this mean that mimeType above can be `YAML` and `YML`? These should be normalized to YAML early.

timj (approved):
Looks okay. I do have come suggestions and questions.

python/HeaderService/hslib.py
menanteau: For now this is only and internal format name.

python/HeaderService/hutils.py
menanteau: Actually you are right, `YML` is redundant at this point as this get normalized at an earlier stage. Will fix it.

python/HeaderService/hutils.py
menanteau: No need really. I changed to simply `yaml_header`

python/HeaderService/hslib.py
timj: I'd call it `fileType` or something then. MIME type implies something that you are not giving.

menanteau: That is the definition for a LFO (Large File Object) message from T&S, not mine. If you prefer we could use a proper mimeType like `text/yaml` for yaml. I kind of prefer this last option, but maybe Archiver doesn't like the extra complexity.

timj: Eek. Calling something a MIME type that is not a MIME type seems odd to me. YAML is generally `application/vnd.yaml` but a lot of people use `text/yaml`. FITS is `application/fits`.

menanteau: I like this. I'll write a little function to assign `application/fits` to FITS mode and `text/yaml` for YAML mode.

PR #34, opened 2019-08-23
DM-21044: AuxTel header updates

menanteau:
First batch of AuxTel header updates

### DIFF
--- a/etc/ATSCam/primary_hdu.header
+++ b/etc/ATSCam/primary_hdu.header
@@ -51,7 +51,7 @@ INSTPORT=                      / The instrument port where LATISS is installed
 ATM3PORT=                      / The port the M3 is directing light into
 TRACKSYS=                      / Tracking coordinate system from ATMCS
 RADESYS =                      / Equatorial coordinate system
-FOCUSZ  =                      / Focus Z position from ATHexapod (mm)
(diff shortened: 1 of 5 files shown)


etc/ATSCam/primary_hdu.header
```diff
@@ -47,6 +47,11 @@ HAEND   =                      / [HH:MM:SS] Telescope hour angle at end
 ELEND   =                      / [deg] Telescope zenith distance at end
 AZEND   =                      / [deg] Telescope azimuth angle at end
 AMEND   =                      / Airmass at end
+INSTPORT=                      / The instrument port where LATISS is installed
+ATM3PORT=                      / The port the M3 is directing light into
+TRACKSYS=                      / Tracking coordinate system from ATMCS
+RADESYS =                      / Equatorial coordinate system
+FOCUSZ  =                      / Focus Z position from ATHexapod (mm)
```
patrickingraham: So at the moment this comes from the ATHexapod, but in the future it'll come from the ATAOS. Maybe say, "Focus position from ATAOS/ATHexapod (mm)"

patrickingraham (approved):
Looks good, just one (very) minor fix.

etc/ATSCam/primary_hdu.header
menanteau: Thanks, will update accordingly.

PR #35, opened 2019-08-23
DM-20156: Adjust travis flake8 script to include bin scripts

menanteau:


.travis.yml
```diff
@@ -5,4 +5,6 @@ matrix:
     - python: '3.6'
       install:
       - pip install flake8
-      script: flake8
+      script:
+      - flake8
+      - flake8 `find bin -type f | while read in ; do if file --mime-type "${in}" | grep -q python ; then echo "${in}" ; fi ; done;`
```
timj: This is more convoluted than I think we really need. The while loop and if statements make it a bit hard to follow. Something like this seems to be cleaner and shorter and gives the same answer:
```
find bin -type f -exec file --mime-type {} \; | grep python | cut -f1 -d:
```
oh, and `$()` syntax is preferred over backticks these days I think.

timj (commented):
I have one substantive comment about the file detection.

Please fix the commit messages to be more accurate. "Test travis changes" would be more usefully written as "Include flake8 test of bin scripts in travis".

Also please remember to have a useful PR title rather than just the branch name.

timj (approved):
Okay to merge once you've squashed and fixed the commit messages.

.travis.yml
menanteau: Thanks for the ninja trick -- updated.

PR #37, opened 2019-09-06
Add await to Remote aths.start_task

menanteau:


bin/csc_send_to_state
```diff
@@ -19,6 +20,10 @@ def cmdline():
 async def main(args):
     async with salobj.Domain() as domain:
         aths = salobj.Remote(domain=domain, name=args.csc_name, index=0)
+        print(f"Waiting for {args.csc_name} to complete Start")
```
r-owen: This is fine but I have a few small suggestions for improvement:
1) Skip the time measurement. Remote already logs a message saying how long it waited for historical data, and that is essentially the same thing you are measuring and reporting.
2) Consider using Remote as an async context manager, since you can in this situation. That avoids the need to await the start task. (I have also taken advantage of 0 being the default index):

```
async with salobj>Domain() as domain, \
        salobj.Remote(domain=domain, name=args.csc_name) as remote:
    newstate = ...
```

r-owen (approved):
Looks good. One small optional suggested change.

bin/csc_send_to_state
menanteau: Thanks for the suggestion. How can I access the message saying how long it waited for historical data?

r-owen: Good question. I think you'll want to add a logging.StreamHandler to the log in `remote.sal_info`, which is something like:
```
remote.sal_info.log.addHandler(logging.StreamHandler())
```
Not much savings in lines of code, but you might appreciate seeing the other logging messages.

r-owen: And if you still don't see the message then explicitly call `setLevel(logging.INFO)` on the log, but I'm pretty sure that is the default for salobj.

menanteau: I cannot get it to print the log. It works but no messages are printed. This is what I'm doing:
```
async def main(args):
    async with salobj.Domain() as domain, \
            salobj.Remote(domain=domain, name=args.csc_name) as remote:
        remote.salinfo.log.addHandler(logging.StreamHandler())
        remote.salinfo.log.setLevel(logging.INFO)
        newstate = getattr(salobj.State, args.state)
        remote.salinfo.log.info(f"Sending {args.csc_name} to: {args.state}")
        await salobj.set_summary_state(remote=remote, state=newstate)
        remote.salinfo.log.info(f"Done")
```

r-owen: That is very strange. This simple code works for me which makes me think the above should work for you:
```
import logging

log = logging.getLogger("test")
log.info("I do not see this message")
# set level before adding the handler to match what salobj does
# but it should work either way
log.setLevel(logging.INFO)
log.addHandler(logging.StreamHandler())
log.info("I see this message")
```

PR #38, opened 2019-09-16
Add basic WCS-TAN to AuxTel headers

menanteau:


etc/ATSCam/primary_hdu.header
```diff
@@ -63,8 +80,8 @@ GROUPID =                      /
         ---- Information from Camera
 CCD_MANU= 'ITL'                / CCD Manufacturer
 CCD_TYPE= '3800C'              / CCD Model Number
-CCD_SERN= '20304'              / Manufacturers? CCD Serial Number
-LSST_NUM= 'ITL-3800C-098'      / LSST Assigned CCD Number
+CCD_SERN= '20862'              / Manufacturers? CCD Serial Number
+LSST_NUM= 'ITL-3800C-068'      / LSST Assigned CCD Number
```
RobertLuptonTheGood: Two problems:  why is this change in a WCS ticket, and why are we hard-coding this information in the header service?  The value is known in cameraGeom and presumably to the camera;  if you are not getting it from there please omit it.

etc/ATSCam/segment_hdu.header
```diff
@@ -12,3 +12,29 @@ DTM1_1  =                      / detector transformation matrix
...
+CRPIX1A =
+CRPIX2A =
+CRVAL1A =
+CRVAL2A =
+WCSNAMEC= 'CCD'                / Name of coordinate system
+CTYPE1C = 'CCD_X'              / In the camera coordinate system
+CTYPE2C = 'CCD_Y'              / In the camera coordinate system
+PC1_1C  =
+PC1_2C  =
+PC2_1C  =
```
RobertLuptonTheGood: I don't think we should be writing per-segment WCS information.  It's implied by the whole-CCD Wcs, and the information required to go from amp to CCD geometry is in `cameraGeom` so should not be repeated here.

python/HeaderService/camera_coords.py
```diff
@@ -253,62 +291,170 @@ def mosaic_ITL(self, Sx, Sy):
...
         return wcs
 
-    def wcs_CCD(self, Sx, Sy):
+    def wcs_C(self, Sx, Sy):
+
+        """
+        This function defines the WCS representation for a Segment (Sx,Sy) in
+        CCD coordinates as described in section 9.6 of LCA-13501 for e2V and
+        ITL sensors.
+        """
```
RobertLuptonTheGood: LCA-13501 refers to camera-side EOtest, it does not apply to how we will be writing LSST data for science processing.

RobertLuptonTheGood (changes requested):
I've made some notes on the PR, but look at the Jira ticket.  I'd like to see most of this information removed from the headerservice as it's not normative and I don't think you're getting it from the camera.

etc/ATSCam/primary_hdu.header
timj: LSE-400 requires it be there. Also, CCS files seem to include it so the camera has some idea of the LSST_NUM. I agree that hard-coding it in the header service seems wrong if CCS already knows about it. We do want the information in there though since it's the only place in the files that tells us which CCD is really being used (granted, not if it's hard-coded here). All LSST data files have always had LSST_NUM specified.

RobertLuptonTheGood: OK, and I agree that it's good to have it.  However, hard-coding the value is worse than useless, so please get it from the CCS.  If this isn't possible, please open a CAP ticket and we can discuss solutions.

etc/ATSCam/segment_hdu.header
timj: The main question is whether anyone ever needs to do anything with the individual amplifiers without any ISR being applied. ISR is only applied on the assembled data so if you can only ever visualize that then WCS on the amplifier doesn't help. If you want to visualize the segments as a collection in something like DS9 then you do need some form of WCS.

RobertLuptonTheGood: There is a niche case, but running LSST code easily assembles the segments even without running full ISR, and has utilities to give the (ra, dec) for a position in amp coordinates.  I think that the danger of all this extra information embedded in the header service outweighs the advantages

etc/ATSCam/primary_hdu.header
menanteau: According to Tony J these should not change (unless something really bad happens). However, I have already requested Tony to publish these as `ATCamera` events. These will be available via the `ATCamera_logevent_wrebSettingsApplied` event on the next tag of `ts_xml`, which will be probably released this week.

etc/ATSCam/segment_hdu.header
menanteau: I am agnostic in regards to adding the `PCX_XXC` matrix. However Tony made the request to include these keywords as the Camera software uses it for visualization and that's why I added in the spirit of making the files useful for most people. Please tell me how you would like me to proceed.

python/HeaderService/camera_coords.py
menanteau: LCA-13501 is relevant here as `ATCamera` is using  this as the reference for the `DTVX/DTMXX` matrix as well as `DATASEC`, `DETSIZE`  and `DETSEC` for each segment. Please let know how you want me to proceed.

etc/ATSCam/segment_hdu.header
menanteau: This has been removed now.

etc/ATSCam/primary_hdu.header
menanteau: The `master` branch of `ATHeaderService` now has the above changes as part of the `ts_xml 4.3.0` updates on commit `25c2ca1`

RobertLuptonTheGood: I'm confused.  Your comment says that the `ts_xml 4.3.0` is broadcasting the CCD serial number, but you're still hard-coding it here.  Also, could you explain why this change is in a WCS ticket?

python/HeaderService/camera_coords.py
```diff
@@ -57,13 +57,33 @@
...
+                    'ccdax': 4004,
+                    'ccday': 4096,
+                    'ccdpx': 4197,
+                    'ccdpy': 4200,
+                    'gap_inx': 28,
+                    'gap_iny': 25,
+                    'gap_outx': 26.5,
+                    'gap_outy': 26,
+                    'raftx': 12700,
+                    'rafty': 12700}
```
RobertLuptonTheGood, line 71: Why are these numbers hard-coded here, and why are there numbers for both e2v and itl devices (you just argued that the CCD serial number isn't going to change).  I am not happy that any of these numbers are in the header service -- LSST gets them from the cameraGeometry and they are not reliable even for the telescope as the number of overscan pixels can change when the sequencer files are changed,

python/HeaderService/camera_coords.py
```diff
@@ -162,14 +182,25 @@ def __init__(self, vendor,
...
+        self.ccday = SCAN_GEOM[vendor]['ccday']
 
         # Make this info available as part of the class
         self.CHANNEL = CHANNEL[vendor]
         self.SEGNAME = SEGNAME[vendor]
 
+        # values needed for WCS
+        self.CDELT1 = xpixelscale
+        self.CDELT2 = ypixelscale
+        # if undefined, we assume reference position it's in the center
```
RobertLuptonTheGood: typo: "is in the"

python/HeaderService/camera_coords.py
RobertLuptonTheGood: I don't understand.  What do you mean by, "ATCamera" -- the CCS, in which case I'm confused as this is being written in the "OCS" side?  None of the DM code needs these numbers.

python/HeaderService/camera_coords.py
```diff
@@ -253,62 +287,170 @@ def mosaic_ITL(self, Sx, Sy):
...
+                   'PC2_1C': -1,
+                   'PC2_2C': 0,
+                   'CDELT1C': 1,
+                   'CDELT2C': 1,
+                   'CRPIX1C': 0,
+                   'CRPIX2C': 0,
+                   'CRVAL1C': Sx*(2*self.dimv+1),
+                   'CRVAL2C': self.dimh + 1 + Sy*self.dimh - self.preh}
+        return wcs
+
```
RobertLuptonTheGood: Are you reading the vendor from CCS (see above)?
Why are we writing a "C" WCS?
Why are we using `PCn_m` keys here and `CDn_m` earlier; we should use one or the other consistently.  The standard forbids us from mixing `CD` and `PC` cards in a single HDU, and while it's probably OK this formally forbids us to mix-and-match even with `WCS` and `WCSC`

python/HeaderService/camera_coords.py
```diff
@@ -253,62 +287,170 @@ def mosaic_ITL(self, Sx, Sy):
...
+        yc: float
+            The y value on CCD coordinates.
+        """
+
+        if self.vendor == 'E2V':
+            yc = xs*(1-2*Sx)+Sx*(self.dimh+1)+Sy*self.dimh+(2*Sx-1)*self.preh
+        elif self.vendor == 'ITL':
+            yc = 1 - xs + self.dimh*(1 + Sy) - self.preh
+        xc = (1 - 2*Sx)*ys + Sx*(2*self.dimv + 1)
+        return xc, yc
```
RobertLuptonTheGood: I thought we'd decided to remove per-segment WCS entries, as agreed with Tony.  What's going on here?

python/HeaderService/hslib_salobj.py
```diff
@@ -427,7 +427,23 @@ def collect_from_HeaderService(self):
...
         else:
-            self.log.info("No 'ELEND' and 'AZEND'")
+            self.log.info("No 'ELEND' and 'AZEND' to compute RA/DEC END")
+
+        if 'RA' and 'DEC' and 'ROTPA' in self.metadata:
+            # --- TODO ----
+            # In the near future, once radesys comes from telemetry,
+            # we should use self.metadata['RADESYS'] instead of the
+            # hardcorded FK5 value we pluged here.
+            # --------------
```
RobertLuptonTheGood: What's the ticket number for adding `RADECSYS`?

RobertLuptonTheGood (changes requested):
Sorry for the delay.

etc/ATSCam/primary_hdu.header
menanteau: I've merged `master` into `tickets/DM-19669` to reflect the changes that were made a while on the main branch. The values are no longer hard coded.

python/HeaderService/camera_coords.py
menanteau: The `ATHeaderService` needs these values which are captured from the `ATCamera.imageReadoutParameters`. These are broadcasteded by `ATCamera` (i.e. CCS) and the `ATHeaderService` is agnostic about where is `ATCamera` getting them from.

python/HeaderService/camera_coords.py
menanteau: Although these are not used, I've keep the functions in case Tony wants this back in the future. If you want me to removed them, say so, and I will.

python/HeaderService/camera_coords.py
menanteau: Same as above, we are not using this function. I kept in the code in case we want to use it in the future. If you want it removed, just say remove it.

python/HeaderService/camera_coords.py
menanteau: Fixed.

python/HeaderService/camera_coords.py
menanteau: I added them to have access to these definitions in the future. If you want them out, I can removed them.

python/HeaderService/hslib_salobj.py
menanteau: Couldn't find it with my jira knowledge. Will keep trying.

menanteau: https://github.com/lsst-dm/HeaderService/pull/34/files

timj:
@menanteau why is there a merge commit from master on to this ticket branch? That's not DM development policy so I'm a bit confused. (commit b6a120472bdef7a25977b41e3343afea78d74858). DM policy is to rebase.

menanteau:
@timj I made this so Robert @RobertLuptonTheGood could see the changes that I've made to `master` regarding `CCD_SERN` and `LSST_NUN`.

timj:
@menanteau to clarify, you are not going to merge with that commit still in place?

menanteau:
@timj No, I will rebase before. This is for @RobertLuptonTheGood to see.

timj:
I think @RobertLuptonTheGood would prefer a proper rebase and some squashing of irrelevant commits so that I can see what this PR is really changing.

menanteau:
@timj I just did that.

timj:
Perfect. I was thrown by going out to lunch and then seeing that this ticket is out of date.

python/HeaderService/hslib_salobj.py
```diff
@@ -475,7 +475,23 @@ def collect_from_HeaderService(self):
             self.metadata['RAEND'] = ra
             self.metadata['DECEND'] = dec
         else:
-            self.log.info("No 'ELEND' and 'AZEND'")
+            self.log.info("No 'ELEND' and 'AZEND' to compute RA/DEC END")
+
+        if 'RA' and 'DEC' and 'ROTPA' in self.metadata:
```
timj: I don't think this is doing what you expect. I think it's only checking that `ROTPA` is in `self.metadata`. The `RA` and `DEC` are `True` independently. I think you either need three separate tests or you need to extract the keys from self.metadata as a set and ask if the `{"RA", "DEC", "ROTPA"}` set is in that set (subset relationship).  This construct is used elsewhere.

python/HeaderService/camera_coords.py
RobertLuptonTheGood: I'd like all non-essential code removed from this changeset, please.  In other words, a minimal changeset that adds a minimal WCS.

python/HeaderService/camera_coords.py
```diff
@@ -162,14 +182,25 @@ def __init__(self, vendor,
...
         # Make this info available as part of the class
         self.CHANNEL = CHANNEL[vendor]
         self.SEGNAME = SEGNAME[vendor]
 
+        # values needed for WCS
+        self.CDELT1 = xpixelscale
+        self.CDELT2 = ypixelscale
+        # if undefined, we assume reference position is in the center
+        # of the sensor.
+        if not crpix1:
```
timj: Since crpix1 can have a valid value of 0.0, which is False, I think this code should say
```python
if crpix1 is None:
    ...
```
Same below as well.

python/HeaderService/camera_coords.py
```diff
@@ -138,11 +158,11 @@ def __init__(self, vendor,
                  overv=None,
                  xpixelscale=0.105,
                  ypixelscale=0.105,
+                 crpix1=None,
+                 crpix2=None,
                  ):
         self.vendor = vendor
```
timj, line 148: I think that since you are doing:
```python
if self.vendor == "ITL":
    ...
elif self.vendor == "E2V"
   ...
```
without having an `else` I think it would be a good idea to validate `vendor` here to ensure that you know things have gone wrong early.

menanteau: The check for the allowed `vendor` is done when the function `HDR.load_templates()` is called. This happens at startup and when a new image is triggered. At the moment the type of vendor is set up by the HeaderService config file at startup (`config.vendor`). However, after the last `ts_xml` update that information is now available from `ATCamera` via the event `ATCamera.wrebSettingsApplied.ccdManufacturer`. I think we should change the current implementation and grab the info from the event instead, as this would be the need for LSSTCam too. I would like to do this on a separate ticket as it doesn't relate to the WCS.

python/HeaderService/camera_coords.py
menanteau: Thanks I've fixed that.

python/HeaderService/hslib_salobj.py
menanteau: Thanks for catching that. It's been fixed.

python/HeaderService/hslib_salobj.py
```diff
@@ -475,7 +475,23 @@ def collect_from_HeaderService(self):
...
+            # --- TODO ----
+            # In the near future, once radesys comes from telemetry,
+            # we should use self.metadata['RADESYS'] instead of the
+            # hardcorded FK5 value we pluged here.
+            # --------------
+            self.log.info("Computing WCS-TAN from RA/DEC/ROTPA")
+            wcs_TAN = self.HDR.CCDGEOM.wcs_TAN(self.metadata['RA'],
+                                               self.metadata['DEC'],
+                                               self.metadata['ROTPA'],
+                                               'FK5')
```
timj: ICRS should be a better default shouldn't it?

menanteau: I though Fk5 was, but I'm happy to change it to ICRS if you think we should.

timj: It of course depends on the system used for RA DEC but I thought we were using ICRS. FK5 is an old system that isn't quite ICRS but is close.

python/HeaderService/hslib_salobj.py
```diff
@@ -461,10 +461,10 @@ def collect_from_HeaderService(self):
             self.metadata['RASTART'] = ra
             self.metadata['DECSTART'] = dec
         else:
-            self.log.info("No 'ELSTART' and 'AZSTART'")
+            self.log.info("No 'ELSTART' and 'AZSTART' to compute RA/DEC START")
```
RobertLuptonTheGood: This implies that neither `ELSTART` nor `AZSTART` is in the header.  Please change the message to be specific about which key[s] are missing

python/HeaderService/hslib_salobj.py
```diff
@@ -475,7 +475,23 @@ def collect_from_HeaderService(self):
...
+            # --------------
+            self.log.info("Computing WCS-TAN from RA/DEC/ROTPA")
+            wcs_TAN = self.HDR.CCDGEOM.wcs_TAN(self.metadata['RA'],
+                                               self.metadata['DEC'],
+                                               self.metadata['ROTPA'],
+                                               'FK5')
+            # Update the metadata
+            self.metadata.update(wcs_TAN)
+        else:
+            self.log.info("No 'RA', 'DEC' or 'ROTPA' to compute WCS")
```
RobertLuptonTheGood: Please be specific about which cards are missing

python/HeaderService/hslib_salobj.py
```diff
@@ -475,7 +475,23 @@ def collect_from_HeaderService(self):
             self.metadata['RAEND'] = ra
             self.metadata['DECEND'] = dec
         else:
-            self.log.info("No 'ELEND' and 'AZEND'")
+            self.log.info("No 'ELEND' and 'AZEND' to compute RA/DEC END")
```
RobertLuptonTheGood: Please be specific about which cards are missing

python/HeaderService/hslib_salobj.py
RobertLuptonTheGood: That's not a ticket number.  Is it DM-6576?

RobertLuptonTheGood (commented):
As Felipe and I discussed in person this functionality needs cleanup to remove all hard-coded numbers that could come from the camera (especially ones that are only known at CCD init time, e.g. number of overscan rows/columns).   I'll file a ticket.

In the meantime, this ticket is good to go.

python/HeaderService/hslib_salobj.py
menanteau: Sure, I will update to ICRS.

python/HeaderService/hslib_salobj.py
menanteau: Fixed as requested.

python/HeaderService/hslib_salobj.py
menanteau: Fixed as requested.

python/HeaderService/hslib_salobj.py
menanteau: Since the last review, `RADECSYS` has been included `ATMCS_logevent_target.radecsys` since ts_xml 4.4.1. So, I will use `self.metadata['RADESYS']` that it is capturing `ATMCS_logevent_target.radecsys`. However, like a few of the keywords expected by TSS it might not be populated yet. We will need to continue to audit the header for completeness.

python/HeaderService/hslib_salobj.py
menanteau: Fixed at requested.

RobertLuptonTheGood (approved):

PR #39, opened 2019-09-17
Fix logging missing from sub-modules

menanteau:


python/HeaderService/hutils.py
```diff
@@ -66,6 +61,9 @@ def configure_logger(logger, logfile=None, level=logging.NOTSET, log_format=None
         FORMAT_DATE = '%Y-%m-%d %H:%M:%S'
     formatter = logging.Formatter(FORMAT, FORMAT_DATE)
 
+    # Need to set the root logging level
```
r-owen: Please delete or expand on this comment. It is clear you are setting the log level from the code, which argues for deletion. But if there is some subtlety you are trying to document, please explain more.

python/HeaderService/hutils.py
```diff
@@ -218,33 +216,43 @@ def get_image_size_from_imageReadoutParameters(myData):
...
 
-def start_web_server(dirname, port_number=8000, httpserver="http.server"):
+def start_web_server(dirname, port_number=8000, httpserver="http.server", logger=None):
+
+    if not logger:
+        LOGGER = create_logger()
+        LOGGER.setLevel(logging.INFO)
+        print("Creating new LOGGER")
+    else:
+        LOGGER = logger
```
r-owen: I strongly suggest you replace LOGGER with logger. Names in all caps names should be reserved for global constants (and variables, but don't use those!). It also allows you to simplify the code to the following:
```
if not logger:
    logger = create_logger()
    logger.setLevel(logging.INFO)
    self.log.info("Created a new logger")
```
where I also took the liberty of changing the print statement to a log statement.

python/HeaderService/hutils.py
```diff
@@ -310,22 +326,22 @@ def load_templates(self):
         self.header_primary = read_head_template(self.templ_primary_file)
 
         # Load up the template for the PRIMARY header
-        LOGGER.info("Loading template for: {}".format('PRIMARY'))
+        self.log.info("Loading template for: {}".format('PRIMARY'))
```
r-owen: Consider `self.log.info("Loading template for PRIMARY")`

format is overkill for a fixed string.

python/HeaderService/hutils.py
```diff
@@ -310,22 +326,22 @@ def load_templates(self):
...
         # For the Segments, we load it once and then copy and
         # modify each segment
         for SEG in self.segment_names:
             EXTNAME = '{}{}'.format(self.segname, format(SEG))
-            LOGGER.info("Loading template for: {}".format(EXTNAME))
+            self.log.info("Loading template for: {}".format(EXTNAME))
             self.header[EXTNAME] = copy.deepcopy(self.header_segment)
             # Now get the new value for the SEGMENT
-            LOGGER.debug("Updating GEOM in template for: {}".format(SEG))
+            self.log.debug("Updating GEOM in template for: {}".format(SEG))
```
r-owen: Consider using f strings instead of format, for readability. For example (also avoiding all caps):
```
for seg in self.segment_names:
    extname = self.segname + format(seg)  # or f"{self.segname}{seg}"
    self.log.info(f"Loading template for {extname}")
    self.header[extname] = copy.deepcopy(self.header_segment)
    # Now get the new value for the segment
    self.log.debug(f"Updating GEOM in template for {seg}")
```

r-owen (approved):
This looks fine. I had a few minor suggestions, the main two of which are:
* Please consider replacing your all caps variable names with lowercase so other python programmers have an easier time understanding the code. All caps usually means a global constant. That's an easy set of search/replace.
* Please consider using f strings instead of `format` (you could do this incrementally, e.g. in new code) since they are shorter and easier to read.

python/HeaderService/hutils.py
menanteau: Will replace `LOOGER` by `logger`.
Thanks for catching the leftover `print` statement.

python/HeaderService/hutils.py
menanteau: Let me know if makes sense to expand. If too obvious I'll delete.
The reason for changing the level for the root logging level, is because setting the level for each of the handler won't be recognized unless the root level is set at the appropriate level. For example, if you set the root logger to INFO, and all handlers to DEBUG, you still won't receive DEBUG messages on handlers.

r-owen: Hmm...I think the key is whether you'll need the information yourself if you modify the code later. If so I suggest you expand the comment, otherwise remote it.

python/HeaderService/hutils.py
menanteau: Will do. Thanks.

python/HeaderService/hutils.py
menanteau: Fixed. Thanks.

python/HeaderService/hutils.py
menanteau: Will augment the comment.

PR #41, opened 2019-09-20
Changes due to for ts_xml 4.2.0

menanteau:


bin/telemetry_sim_ATHS
```diff
@@ -56,6 +56,15 @@ def cmdline():
     parser.add_argument("--focusz", action="store", default=13, type=float,
                         help="ATHexapod/ATAOS FocusZ position (mm)")
 
+    parser.add_argument("--instport", action="store", default=1, type=int,
+                        help="The instrument port where LATISS is installed")
+    parser.add_argument("--m3port", action="store", default=1, type=int,
+                        help="Port the M3 is directing light into")
+    parser.add_argument("--tracksys", action="store", default="Local", type=str,
+                        help="The track system value will [Local, Sidereal, or non-sidereal.]")
```
patrickingraham: The tracking coordinate system [Local, Sidereal, or non-sidereal].

bin/telemetry_sim_ATHS
```diff
@@ -110,7 +117,14 @@ if __name__ == "__main__":
             'priority': 1,
             'force_output': True}
     ptg.evt_currentTarget.set_put(**kwTA)
-    LOGGER.info("Sending currentTarget")
+    LOGGER.info("Sending Ptg currentTarget")
```
patrickingraham: What do you mean by "sending" in this context?? It might be better to use the word "Sending simulated Ptg currentTarget event"

(also in line 111 "Poiting" is missing an 'n' :)

patrickingraham: This also applies to all the "sending" strings below as well.

python/HeaderService/hslib_salobj.py
```diff
@@ -377,7 +377,14 @@ def collect(self, keys):
             if self.myData[name] is None:
                 self.log.warning(f"Cannot get keyword: {k} from topic: {name}")
             else:
-                self.metadata[k] = getattr(self.myData[name], param)
+                # In case of a long telemetry array we take the first element
```
patrickingraham, line 380: The main telescope will do something similar, but probably have a (roughly) equal number from what we have with the AT... maybe more due to the hexapod etc...

patrickingraham (commented):
Looks good, recommend altering some log messages for clarity sake.

bin/telemetry_sim_ATHS
menanteau: Thanks, fixed.

bin/telemetry_sim_ATHS
menanteau: Yes, these are all simulated messages. I'll add to the logger.

PR #43, opened 2019-10-25
Update Dockefile and utilities to start it

menanteau:


start_docker_service.sh
```diff
@@ -0,0 +1,24 @@
+HSUSER=headerservice
```
srp3rd: Is there any reason to have to initialize this variable, in this script?

start_docker_dev.sh
```diff
@@ -1,6 +1,7 @@
 HSUSER=headerservice
-TAGNAME=3.10.0-4.3.0-salobj_4.5.0
+TAGNAME=3.10.0-4.4.0-salobj_4.5.0-aths_1.1.0
 DOCKER_IMA=lsstdm/atheaderservice:$TAGNAME
+NAME=aths
```
srp3rd, line 4: Is this variable necessary?  Where is it used?

srp3rd (approved):

start_docker_dev.sh
menanteau: I use it for:
`dockerID=$(docker ps -qa --filter name=$NAME)`
in case I want to nuke a previous one that's running.

start_docker_service.sh
menanteau: Not for this script. I'll clean it up.

PR #46, opened 2020-01-28
Adding scale option and OBJECT to header templates.

menanteau:


etc/ATSCam/primary_hdu.header
```diff
@@ -52,6 +52,7 @@ ATM3PORT=                      / The port the M3 is directing light into
 TRACKSYS=                      / Tracking coordinate system from ATMCS
 RADESYS =                      / Equatorial coordinate system
 FOCUSZ  =                      / Focus Z position from ATAOS/ATHexapod (mm)
+OBJECT  =                      / Name of the observed object
```
timj, line 55: Can we put it up the top of this block above RA please?

timj (approved):
Looks fine. Thanks.

etc/ATSCam/primary_hdu.header
menanteau: Sure.

PR #47, opened 2020-01-30
Adding dynamical timeout using requested exptime

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -109,10 +109,17 @@ def start_collection_event_callback(self, data):
         sys.stdout.flush()
         self.log.info(f"Received: {self.name_start} Event")
 
+        # Get the requested exposure time to estimate the timeout
+        exptime_key = self.config.timeout_keyword
+        self.log.info(f"Collecting key {exptime_key} to set timeout as:"
+                      f"{exptime_key} + {self.config.timeout_exptime}")
```
timj: Log messages conventionally use % syntax and trailing arguments rather than creating strings with embedded values. This is so that we can defer filling in the values until we know the log message will be sent out.  Rather than mentioning the key twice how about:

> Collecting key to set timeout as key {exptime_key} + {...timeout}

python/HeaderService/hslib_salobj.py
```diff
@@ -109,10 +109,17 @@ def start_collection_event_callback(self, data):
...
+        self.collect([exptime_key])
+        timeout = self.metadata[exptime_key] + self.config.timeout_exptime
+        self.log.info(f"Using timeout: {timeout} [s]")
```
timj: Again should be:
```python
self.log.info("Using timeout %s [s]", timeout)
```

timj (approved):
Looks okay with minor comments.

python/HeaderService/hslib_salobj.py
menanteau: Done, thanks.

python/HeaderService/hslib_salobj.py
menanteau: Also fixed.

PR #49, opened 2020-02-11
Update to handle more than one image at a time.

menanteau:


bin/telemetry_sim_ATHS
```diff
@@ -66,11 +66,15 @@ def cmdline():
                         help="The tracking coordinate system [Local, Sidereal, or non-sidereal].")
     parser.add_argument("--radesys", action="store", default="ICRS", type=str,
                         help="The ra/dec track system value.")
-    parser.add_argument("--object", action="store", default="ElGrodo", type=str,
+    parser.add_argument("--object", action="store", default="ElGordo", type=str,
```
timj, line 69: Can you please use a meaningful default here and for imagetag below?

menanteau: "El Gordo" is largest galaxy cluster at high redshift, see: https://en.wikipedia.org/wiki/El_Gordo_(galaxy_cluster)

timj: that may well be but "DEFAULTOBJECT" and "DEFAULTIMAGETAG" would be more obvious (and ImageTag should never be a galaxy).

timj:
@menanteau you asked me to review but I'm not an expert on asyncio so it's probably better to get @r-owen to do the review if possible.

python/HeaderService/hslib_salobj.py
```diff
@@ -60,28 +60,29 @@ def __init__(self, **keys):
...
 
-        # Initialize the timeout task
-        self.end_evt_timeout_task = salobj.make_done_future()
-
     async def close_tasks(self):
         """Close tasks on super and evt timeout"""
         await super().close_tasks()
-        self.end_evt_timeout_task.cancel()
+        # Loop over all task and cancel all of them
+        for imageName in self.end_evt_timeout_task:
```
srp3rd, line 70: You'll need a lock around this code, since you're messing with the end_evt_timeout_task list.

python/HeaderService/hslib_salobj.py
```diff
@@ -60,28 +60,29 @@ def __init__(self, **keys):
...
+        self.clean(imageName)
 
     def report_summary_state(self):
         """Call to report_summary_state and LOGGER"""
         super().report_summary_state()
         self.log.info(f"Current state is: {self.summary_state.name}")
         if self.summary_state != salobj.State.ENABLED:
-            self.end_evt_timeout_task.cancel()
+            # Loop over all task and cancel all of them
+            for imageName in self.end_evt_timeout_task:
```
srp3rd: You'll need a lock around this code, since you're messing with the end_evt_timeout_task list and since you released the lock in the calling method before this method was called.

srp3rd (changes requested):
Just two issues in the code, listed below for the locks.

python/HeaderService/hslib_salobj.py
r-owen: If the code does not use threads then I think it is not necessary to use locks here because the for loop never releases the event loop.

srp3rd: Felipe and I were discussing this in his office while you were making this comment, so that was good timing.  

I've been thinking in a more threading model of things, rather than co-operative tasks.    r-owen is right.  If this had been actual threading, then the lock would have been required at this point, and in the close_tasks() method above it because there would be a race condition.  Since these are co-operative tasks, the race condition doesn't exist.

We had been talking for the more global lock that protects the other data structures;  those probably don't need to be protected, because they never release the event loop.

r-owen: Once the code gets past `await super().close_tasks()` it does not relinquish the event loop again. (Neither `self.end_evt_timeout_task.cancel()` nor the for loop that follows releases the event loop). So I think a lock should not be needed and you run the risk of a deadlock if you use the lock *because* the code does not give up the event loop, so the other code that has the lock cannot give it up.

Resource locks are rarely needed in cooperative multitasking code. I do find them necessary once in awhile. For instance in code that issues a command and waits for a reply on a TCP/IP port I will lock that code, so that another task cannot send a command while the first one is waiting for its reply.

menanteau: Thanks @spr and @r-owen for the comments and the time to check the code.

PR #50, opened 2020-02-12
Add a FACILITY to header template

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -60,28 +60,33 @@ def __init__(self, **keys):
...
+        # Loop over all task and cancel all of them
+        for imageName in self.end_evt_timeout_task:
+            self.end_evt_timeout_task[imageName].cancel()
 
-    async def end_evt_timeout(self, timeout):
+    async def end_evt_timeout(self, imageName, timeout):
         """Timeout timer for end event telemetry callback"""
         await asyncio.sleep(timeout)
-        self.log.info(f"{self.name_end} not seen in {timeout} seconds; giving up")
+        self.log.info(f"{self.name_end} for {imageName} not seen in {timeout} [s]; giving up")
```
timj, line 76: As a general comment, logging should not do string interpolation directly. Log messages should defer interpolation until the system knows the message is going to be issued. For example:
```python
logger.warning("Houston, we have a %s", "bit of a problem")
```
Generally it doesn't really matter but sometimes `__str__` has overhead and it's good to not trigger that unless the message is being issued.

menanteau: I do not have a preference on the formatting, but I already changed all of the logging to the current format as suggested/requester @r-owen

timj: It's nothing to do with f-strings vs %-formatting, it's to do with always deferring formatting in log messages until the system knows that a log message will actually be issued.

menanteau: Do you want to make a ticket to change the logging formatting  to be always deferring across the code?

r-owen: For occasional log messages, simple formatting and all `warning` or higher level messages, I strongly prefer f strings, for two reasons:
* The code is always exercised, avoiding deferred formatting errors.
* Improved code clarity

I agree that using deferred formatting is the right thing to do for `debug` level log messages that are output frequently (e.g. in a loop) or have complicated formatting. At that point performance may be an issue and `debug` messages are rarely shown.

If deferred formatting is used, then the unit tests should enable the necessary log level, in order to detect formatting errors.

timj: Agreed that if we always expect info messages to be issued then f-strings are fine.

PR #54, opened 2020-03-25
Updates for accomodate multi-sensor headers for ComCam and LSSTCam

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -468,7 +478,7 @@ def create_dicts(self):
         self.filename_FITS = {}
         self.filename_HDR = {}
 
-    def collect(self, keys):
+    def collect_old(self, keys):
```
srp3rd: Since the "collect_old" method is no longer in use, it should be deleted.  If you need to look up something it did later, it can be looked up in older versions of this file via git.

start_docker_service.sh
```diff
@@ -1,4 +1,4 @@
-TAGNAME=4.0.0-4.6.0-salobj_5.0.0-aths_1.4.0
+TAGNAME=4.0.0-4.7.0-salobj_5.4.0-aths_1.4.0
 DOCKER_IMA=lsstdm/atheaderservice:$TAGNAME
 NAME=aths
```
srp3rd, line 3: If you're renaming everything else from ATHeaderService to HeaderService, you might want to rename these too.

srp3rd (changes requested):
Only comments are to get rid of a method that's no longer used and to lose the "AT" part of the name since it's been removed elsewhere.  See comments.

python/HeaderService/hslib_salobj.py
menanteau: Thanks for noticing this. I though I have cleaned up all of them....

start_docker_service.sh
menanteau: Thanks for finding this one too.  The `lsstdm/atheaderservice` is a reserved space on the DM dockerhub that I'll need to have Josh change for me to ``lsstdm/headerservice`

menanteau: Got the new name from Josh `lsstdm/headerservice` for dockerhub

PR #55, opened 2020-04-14
Updates from new xml and Camera updates

menanteau:


mareuter (approved):
Looks fine to me. The one small thing is you need to fix the docstring in the announce function in hslib_salobj.py by dropping the "and optionally emulate it for the EFD" since that is no longer done.

PR #56, opened 2020-04-27
Updates for new T&S Telemetry in ts_xml 5.0

menanteau:


start_docker_dev.sh
```diff
@@ -1,5 +1,5 @@
 HSUSER=headerservice
-TAGNAME=4.0.0-5.0.0-salobj_5.8.0-aths_2.0.0
+TAGNAME=4.1.0-5.0.0-salobj_5.10.0-aths_2.0.0
```
mareuter: Is the ATHS tag version updated just before a new release is done?

start_docker_service.sh
```diff
@@ -1,4 +1,4 @@
-TAGNAME=4.0.0-5.0.0-salobj_5.8.0-aths_2.0.0
+TAGNAME=4.1.0-5.0.0-salobj_5.10.0-aths_2.0.0
```
mareuter: Ditto as above.

start_docker_dev.sh
menanteau: Yes, usually after the PR is done. But I should change `aths` to just `hs` or `headerservice`

menanteau:
@mareuter I've update the future version as 2.0.1, which I will tag after I merge to master.
Thanks

mareuter (approved):
Looks good to me. Thanks for the updates.

PR #58, opened 2020-05-12
Update CCHeaderService after updates to CCCamera XML

menanteau:


bin/telemetry_sim_ComCam
```diff
@@ -69,11 +69,16 @@ def cmdline():
...
-    parser.add_argument("--imagetag", action="store", default="ElGordo", type=str,
+    parser.add_argument("--imagetag", action="store", default="TestTag", type=str,
                         help="The imagetag of the observation.")
     parser.add_argument("--imageName", action="store", default=None, type=str,
                         help="Optional imageName")
 
+    parser.add_argument("--additionalKeys", action="store", default='imageType:groupId:testType',
+                        type=str, help="Optional additionalKeys")
+    parser.add_argument("--additionalValues", action="store", default='BIAS:CALSET_XXXX_YYYY:SIMTEST_ABCD',
+                        type=str, help="Optional additionalKeys")
```
mareuter: Should be "Optional additionalValues"

etc/ComCam/primary_hdu.header
```diff
@@ -61,6 +61,7 @@ CONTRLLR=                      / The controller (e.g. O for OCS, C for (A)CCS)
 DAYOBS  =                      / The observation day as defined by image name
 SEQNUM  =                      / The sequence number from the image name
 GROUPID =                      /
+TESTTYPE=                      / Test Suite Type: DARK:FLAT:OBS:PPUMP:QE:SFLAT
```
mareuter: BIAS can also be included here.

python/HeaderService/hslib_salobj.py
```diff
@@ -505,47 +508,54 @@ def extract_from_myData(self, keyword, myData, sep=":"):
             extracted_payload = payload
         # Case 2 -- array of values per sensor
         elif self.config.telemetry[keyword]['array'] == 'CCD_array':
-            self.log.debug(f"{keyword} is an array")
-            ccdnames = self.get_CCD_keywords(keyword, myData, sep)
-            self.log.info(f"For {keyword} extracted ccdnames: {ccdnames}")
+            self.log.debug(f"{keyword} is and array: CCD_array")
```
mareuter: Should be "an array"

python/HeaderService/hslib_salobj.py
```diff
@@ -505,47 +508,54 @@ def extract_from_myData(self, keyword, myData, sep=":"):
...
         elif self.config.telemetry[keyword]['array'] == 'CCD_array_str':
-            self.log.debug(f"{keyword} is string an array")
-            ccdnames = self.get_CCD_keywords(keyword, myData, sep)
-            self.log.info(f"For {keyword} extracted ccdnames: {ccdnames}")
+            self.log.debug(f"{keyword} is string array: CCD_array_str")
+            ccdnames = self.get_array_keys(keyword, myData, sep)
             # Split the payload into an array of strings
             extracted_payload = dict(zip(ccdnames, payload.split(sep)))
+        elif self.config.telemetry[keyword]['array'] == 'indexed_array':
+            self.log.debug(f"{keyword} is and array: indexed_array")
```
mareuter: Should be "an array"

python/HeaderService/hslib_salobj.py
```diff
@@ -505,47 +508,54 @@ def extract_from_myData(self, keyword, myData, sep=":"):
...
         # If some kind of array
         elif hasattr(payload, "__len__") and not isinstance(payload, str):
-            if 'array_index' in self.config.telemetry[keyword]:
-                index = self.config.telemetry[keyword]['array_index']
-                # Extract the requested index
-                extracted_payload = payload[index]
-            else:
-                # Otherwise take first element
-                extracted_payload = payload[0]
+            self.log.debug(f"{keyword} is just an array")
```
mareuter, line 532: These three lines seem to imply a control block that looks to be missing. Please double check this.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [flake8]
 max-line-length = 110
 max-doc-length = 79
-ignore = E133, E226, E228, N802, N803, N806, N812, N815, N816, W504
+ignore = E133, E226, E228, N802, N803, N806, N812, N815, N816, W504, F541, E741
```
mareuter: You cleaned up a number of issues related to F541 so I don't see the need in ignoring it. E741 is a good thing to have and I don't see any instances in your code. As these are not on the curated list, any lines tripping these errors are supposed to be handled per this section **Lines that intentionally deviate from DM’s PEP 8 MUST include a noqa comment** found here: https://developer.lsst.io/python/style.html#code-may-be-validated-with-flake8.

mareuter (changes requested):
Got a couple of things I'd like to see addressed before approving: possible missing control block and ignoring flake8 errors. The other things are minor.

bin/telemetry_sim_ComCam
menanteau: Thanks for catching this.
Fixed.

etc/ComCam/primary_hdu.header
menanteau: Fixed.

python/HeaderService/hslib_salobj.py
menanteau: Thanks for catching my typos.

python/HeaderService/hslib_salobj.py
menanteau: Thanks for catching my typos.

python/HeaderService/hslib_salobj.py
menanteau: It looks like that from the re-structuring of the function, but it's okay.

setup.cfg
menanteau: Sorry, that was for testing only, and I forgot to put them back. I've done it now.

python/HeaderService/hslib_salobj.py
menanteau: This is okay. It looks like something might be missing because of re structuring I did for the function.

menanteau:
@mareuter I believe the changes are ready for you to review again.

mareuter (approved):
I think we are good now. PR approved.

PR #59, opened 2020-05-20
Update to AT telemetry for FOCUSZ

menanteau:


etc/conf/atTelemetry.yaml
```diff
@@ -55,6 +52,7 @@ end_collection_event:
 imageParam_event:
     device: ATCamera
     topic: imageReadoutParameters
+    array_keys: ccdNames
```
patrickingraham, line 55: what's this??

etc/conf/ccTelemetry.yaml
```diff
@@ -53,6 +53,7 @@ end_collection_event:
 imageParam_event:
     device: CCCamera
     topic: imageReadoutParameters
+    array_keys: ccdLocation
```
patrickingraham, line 56: and this...

start_docker_dev.sh
```diff
@@ -1,5 +1,6 @@
 HSUSER=headerservice
 TAGNAME=4.1.1-5.1.0-salobj_5.10.0-hs_2.2.0
```
patrickingraham: should this not be removed if you re-declare right below?

etc/conf/ccTelemetry.yaml
patrickingraham: Does Tony need to review this as well?

patrickingraham (changes requested):
See inline comments. 
There are comcam changes in here, is someone else reviewing those?

etc/conf/atTelemetry.yaml
menanteau: The key that maps to the arrays for CCD/Sensors. Here I only made `array_keys` visible at the configuration label.

etc/conf/ccTelemetry.yaml
menanteau: Same. No, Tony doesn't need to review this.

start_docker_dev.sh
menanteau: Yes, I'll clean this one up.

patrickingraham (approved):
Looks good. Consider rebasing/squashing commits to remove blank line etc.

PR #60, opened 2020-06-05
Updates to setup.py to include header templates for ComCam/LATISS

menanteau:


start_docker_dev.sh
```diff
@@ -1,5 +1,5 @@
 HSUSER=headerservice
-TAGNAME=4.1.2_408-5.1.0-salobj_5.11.0-hs_2.2.0
+TAGNAME=4.1.1-5.1.0-salobj_5.14.0-hs_2.2.0
```
mareuter: Does the HS version here also get updated since you updated other versions or after the release?

mareuter (approved):
Small comment, but looks good to me.

start_docker_dev.sh
menanteau: yes, after it's tagged.

PR #66, opened 2020-07-20
Add TRACKSYS/RADESYS for ComCam and enum type in telemetry configuarion.

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -531,6 +547,11 @@ def extract_from_myData(self, keyword, myData, sep=":"):
             key = self.config.telemetry[keyword]['array_keyname']
             # Extract only the requested key from the dictionary
             extracted_payload = dict(zip(keywords, split_esc(payload, sep)))[key]
+        # Case 3 -- enumeration using idl libraries
+        elif self.config.telemetry[keyword]['array'] == 'enum':
+            Device = self.config.telemetry[keyword]['device']
```
mareuter: Capital case is reserved for classes. Use lower case for this.

mareuter (approved):
One comment needs addressing, but doesn't require a re-review. Otherwise, it looks good to me.

python/HeaderService/hslib_salobj.py
menanteau: Fixed.

PR #68, opened 2020-08-26
Adding S3 capability of the HeaderService

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -155,12 +155,75 @@ def get_ip(self):
...
+        address = socket.getfqdn()
+        if address.find('.ncsa.') >= 0:
+            s3instance = 'nts'
+        elif address.find('tuc') >= 0:
+            s3instance = 'tuc'
+        elif address.find('.cp.') >= 0:
+            s3instance = 'cp'
+        else:
+            s3instance = 'dummy'
+        self.log.info(f"Will use auto-config s3instance: {s3instance}")
```
srp3rd, line 184: I like that the environment variable S3INSTANCE is used to configure this, and it might be best to require that it be set, and fail otherwise.  If in the future the domain name at the base is aliased or changes for some reason and "cp" isn't there, then it could default to "dummy", which could lead to some confusion down the road.

srp3rd (approved):

python/HeaderService/hslib_salobj.py
menanteau: I don't think we have a clear policy on how the s3instances would work yet. So far I've left it pretty flexible, as the s3instance can in addition be defined in the configuration file and on the command-line with `--s3instance`  option.

etc/dev/s3_upload_example.py
```diff
@@ -0,0 +1,85 @@
...
+    # Now we print the contents
+    for bucket in s3conn.buckets.all():
+        print(f" + {bucket}")
+        for file in bucket.objects.all():
+            print(f"   - {file.key}")
+
+if __name__ == "__main__":
+    # A header file example from ComCam
+    ID = 'CC_O_20200521_000008'
+    asyncio.get_event_loop().run_until_complete(main(ID))
```
mareuter, line 85: In python 3.7 this can be reduced to asyncio.run(main(ID))

python/HeaderService/hslib_salobj.py
```diff
@@ -155,12 +155,75 @@ def get_ip(self):
...
+            s3instance = self.config.s3instance
+            self.log.info(f"Will use s3instance in config: {s3instance}")
+            return s3instance
+        elif 'S3INSTANCE' in os.environ:
+            s3instance = os.environ['S3INSTANCE']
+            self.log.info(f"Will use s3instance from environment: {s3instance}")
+            return s3instance
+
+        # Try to auto-figure out from location
+        self.log.warning("The s3instance was not defined in config")
```
mareuter, line 174: The tree below isn't going to work when the HS runs on k8s. getfqdn returns the pod name and looks nothing like a machine name. We need to think a bit more carefully about how to do this.

bin/headerservice
```diff
@@ -63,6 +63,14 @@ def cmdline():
     parser.add_argument("--hdu_delimiter", action="store", default='END',
                         help="Delimiter for string-based headers")
 
+    # S3/http LFO event
+    parser.add_argument("--lfa_mode", action="store", default='http', type=str.lower,
+                        choices=["s3", "http"],
+                        help="The LFA Mode for the event [s3/http]")
+    parser.add_argument("--s3instance", action="store", default=None,
```
mareuter, line 70: Probably going to need a _ls_ entry for running at the base.

python/HeaderService/hslib_salobj.py
```diff
@@ -155,12 +155,75 @@ def get_ip(self):
...
+        address = socket.getfqdn()
+        if address.find('.ncsa.') >= 0:
+            s3instance = 'nts'
+        elif address.find('tuc') >= 0:
```
mareuter, line 178: Missing the dots. Could be matched to a machine with that name in it.

mareuter (changes requested):
Need to address the self-discovery mechanism when using k8s.

etc/dev/s3_upload_example.py
menanteau: Cool, thanks for that tip.

python/HeaderService/hslib_salobj.py
menanteau: Note that the auto-figure option is a last resource option that uses `socket.getfqdn()` and it is only triggered if **neither** `self.config.s3instance` (from the yaml file) or the env variable `S3INSTANCE` are defined. I'm happy to remove the auto-configure part and just let it fail if neither `S3INSTANCE` or `self.config.s3instance` are not defined.

python/HeaderService/hslib_salobj.py
menanteau: Thanks for noticing.

bin/headerservice
menanteau: Sure, do you have a suggestion?

mareuter: The suggestion is in italics above.

python/HeaderService/hslib_salobj.py
mareuter: Sorry. I guess I was confused on that part. The one issue I have with the config file is that begins to imply that the HS is a configurable CSC which currently it is not. However since the ENVVAR is the backup, I guess I'm less concerned about the configuration file.

bin/headerservice
menanteau: OK -- _ls_  for La Serena, right?

mareuter: Correct. :-)

menanteau: Done

mareuter (approved):
OK. I think I'm satisfied with this for now.

python/HeaderService/hslib_salobj.py
menanteau: Actually, by default the `s3instance` in not defined in the configuration file, so it need to be either set by the command-line argument `--s3instance` or the environment.

PR #71, opened 2020-10-09
Updates to HeaderService for salobj 6

menanteau:


r-owen (approved):
This looks fine.

Two requests for future development:
* Please write unit tests for this code. If they are present, I missed them. (I expect to find them in the `tests` directory). I recommend creating Controller instances for the CSCs whose topics you listen to, then have the test write SAL messages from those in such a way as to trigger the behavior you are testing. In other words: trivial mock CSCs that just write messages at your command.
* Please consider moving the bulk of your bin scripts into library code (in python/HeaderService). That makes it much easier to write unit tests for the code.  (Testing command parsers, for example, is quite easy -- you can feed the arguments directly to the parser, or set sys.arvg and have them picked up that way).

r-owen:
Another small suggestion: in several place you override report_summary_state. Please consider overriding handle_summary_state (a coroutine) instead. Change:
```
def report_summary_state(self):
    super().report_summary_state()
    # your code
```
to:
```
async def handle_summary_state(self):
    # your code, unchanged. Note that there is no need to call `await super().handle_summary_state()`
```

r-owen:
And yet one more suggestion (trivial):

in every subclass of BaseCsc add the following to avoid deprecation warnings (and optionally to support a --version argument):
```
from . import __version__

class HSWorker(salobj.BaseCsc):
    """...doc string..."""

    valid_simulation_modes = [0]    # add this line to avoid a deprecation warnings
    version = __version__  # this line supports the --version command-line argument
```
I admit that you will have to add the `--version` command-line argument yourself, since you aren't using `amain` from `BaseCsc`, but it's an easy addition if you want it. And maybe someday you'll use the built-in `amain` method.

PR #77, opened 2020-11-19
HeaderService updates for LATISS and xml 6.2.1

menanteau:
More updates to LATISS telemetry

Update to ComCam sim script

### DIFF
--- a/bin/telemetry_sim_ComCam
+++ b/bin/telemetry_sim_ComCam
@@ -36,10 +36,6 @@ def cmdline():
     parser.add_argument("--filterpos", action="store", default=666.0,
                         help="The filter measured position of slide (i.e.: 666.0)")
 
-    parser.add_argument("--groupId", action="store", default='LSSTTEST',
-                        help="groupId")
-    parser.add_argument("--imageType", action="store", default='OBJECT',
-                        help="imageType")
     # Control imageName
     parser.add_argument("--telcode", action="store", default='CC',
                         help="The code for the telecope")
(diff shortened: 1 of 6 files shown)


bin/telemetry_sim_ComCam
```diff
@@ -230,7 +225,7 @@ async def amain():
                  'imageName': imageName,
                  'imageIndex': k,
                  'imageNumber': seqN,
-                 'imageSource': 'CC',
+                 'imageSource': args.telcode,
```
mareuter, line 228: This really isn't a telescope code, it's a camera code. ComCam and LSSTCam will both go on MainTel. This goes for the uses in the other script too.

mareuter (approved):
One minor comment that you can do with what you wish. Otherwise, this looks fine to me.

bin/telemetry_sim_ComCam
timj: The problem being that we call this TELCODE explicitly in the FITS header even though it's MC (main camera) for LSSTCam and CC for ComCam. It all got confused by LATISS using AT for AuxTel, because at one point no-one was thinking of AuxTel as a telescope + instrument.

We should consider using CAMCODE as the FITS header instead of TELCODE to fix this. You might want AT to be grandfathered in.

cc/ @RobertLuptonTheGood

menanteau: But this goes into the keyword `TELCODE` hence the name `args.telcode`

menanteau: I see no reason not to change `TELCODE` --> `CAMCODE` for all AT/CC/MT. If that's the case, let me know so I can make a ticket.

RobertLuptonTheGood: Sounds like it's worth following Tim's advice here.

PR #78, opened 2020-11-19
Change TELCODE to CAMCODE for HeaderService

menanteau:


etc/ComCam/primary_hdu.header
```diff
@@ -54,7 +54,7 @@ FOCUSZ  =                      / Focus Z position
 OBJECT  =                      / Name of the observed object
 
         ---- Image-identifying used to build OBS-ID ----
-TELCODE =                      / The code for the telescope (AT/CC/MC)
+CAMCODE =                      / The code for the telescope (AT/CC/MC)
```
mareuter: Should say "code for the camera"

etc/LATISS/primary_hdu.header
```diff
@@ -76,7 +76,7 @@ EQUINOX =             2000.00  / Equinox of coordinates
 RADESYS =                      / Telescope coordinate system
 
         ---- Image-identifying used to build OBS-ID ----
-TELCODE =                      / The code for the telescope (AT/CC/MC)
+CAMCODE =                      / The code for the telescope (AT/CC/MC)
```
mareuter: Ditto from above.

mareuter (approved):
Small comment to address. Otherwise, this looks fine to me.

etc/ComCam/primary_hdu.header
menanteau: Fixed, thanks

etc/LATISS/primary_hdu.header
menanteau: Fixed too.

PR #85, opened 2021-01-13
Updates to time calculation

menanteau:


python/HeaderService/hscalc.py
```diff
@@ -33,7 +33,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def get_date(timeStamp=None, scale='tai'):
+def get_date(timeStamp=None, format='unix', scale='utc'):
```
mareuter, line 36: Looks like you need to add format and scale parameters to the function documentation.

mareuter (approved):
One minor comment. Otherwise it looks good.

python/HeaderService/hscalc.py
menanteau: Thanks, done now,

PR #87, opened 2021-01-26
Added back the scale option that was accidentally removed for ComCam dev.

menanteau:


etc/LATISS/primary_hdu.header
```diff
@@ -40,11 +40,11 @@ RAEND   =                      / RA of telescope from AZEND and ELEND
 DECEND  =                      / DEC of telescope from AZEND and ELEND
 ROTPA   =                      / Rotation angle relative to the sky (deg)
 ROTCOORD= 'sky'                / Telescope Rotation Coordinates
-HASTART =                      / [HH:MM:SS] Telescope hour angle at start
+HASTART =                      / [deg] Telescope hour angle at start
```
mareuter: This is decimal hours not degrees.

etc/LATISS/primary_hdu.header
```diff
@@ -40,11 +40,11 @@ RAEND   =                      / RA of telescope from AZEND and ELEND
...
 ELSTART =                      / [deg] Telescope zenith distance at start
 AZSTART =                      / [deg] Telescope azimuth angle at start
 AMSTART =                      / Airmass at start
-HAEND   =                      / [HH:MM:SS] Telescope hour angle at end
+HAEND   =                      / [deg] Telescope hour angle at end
```
mareuter: This is decimal hours not degrees.

mareuter (commented):
Units aren't right for the comment.

etc/LATISS/primary_hdu.header
menanteau: Thanks for noticing, changed to `[decimal hrs]`

etc/LATISS/primary_hdu.header
menanteau: Thanks for noticing, changed to `[decimal hrs]`

mareuter (approved):
Looks good to me.

etc/LATISS/primary_hdu.header
```diff
@@ -40,11 +40,11 @@ RAEND   =                      / RA of telescope from AZEND and ELEND
 DECEND  =                      / DEC of telescope from AZEND and ELEND
 ROTPA   =                      / Rotation angle relative to the sky (deg)
 ROTCOORD= 'sky'                / Telescope Rotation Coordinates
-HASTART =                      / [HH:MM:SS] Telescope hour angle at start
+HASTART =                      / Telescope hour angle at start [decimal hrs]
```
timj: Units should go at the start of the string in FITS convention (which means that ROTPA is wrong above but ELEND is correct below).

in this case though "decimal hours" is not an official fits units string so it would be best to leave it at the end so tooling doesn't get confused.

The standard says that you should use `[h]` so can you please switch to that?

menanteau: Thanks for catching this. It's been updated.

menanteau: @timj Did you mean change to `[decimal h]` or just `[h]`

mareuter: That sounds good to me.

timj: When I said `[h]` I meant `[h]`.  It's a floating point number in the header so it must be decimal.

etc/LATISS/primary_hdu.header
```diff
@@ -38,21 +38,21 @@ RASTART =                      / RA of telescope from AZSTART and ELSTART
...
-ROTPA   =                      / Rotation angle relative to the sky (deg)
+ROTPA   =                      / [deg] Rotation angle relative to the sky
 ROTCOORD= 'sky'                / Telescope Rotation Coordinates
-HASTART =                      / [HH:MM:SS] Telescope hour angle at start
+HASTART =                      / Telescope hour angle at start [decimal h]
 ELSTART =                      / [deg] Telescope zenith distance at start
 AZSTART =                      / [deg] Telescope azimuth angle at start
 AMSTART =                      / Airmass at start
-HAEND   =                      / [HH:MM:SS] Telescope hour angle at end
+HAEND   =                      / Telescope hour angle at end [decimal h]
```
timj: This is still wrong.

```
HAEND =           / [h] Telescope hour angle at end
```
is what it should be. (with the right spacing)

menanteau: Fixed to 
```
HAEND =           / [h] Telescope hour angle at end
```

PR #90, opened 2021-02-10
Updates to deal with telemetry expected as array but appearing as scalars

menanteau:


etc/LATISS/primary_hdu.header
```diff
@@ -91,24 +91,9 @@ IMAGETAG=                      / DAQ Image id
...
         ---- Information from Camera ----
-CCD_MANU=                      / CCD Manufacturer
 CCD_TYPE=                      / CCD Model Number
-CCD_SERN=                      / Manufacturers CCD Serial Number
-LSST_NUM=                      / LSST Assigned CCD Number
-SEQCKSUM=                      / Checksum of Sequencer
-SEQNAME =                      / SequenceName from Camera StartIntergration
-REBNAME =                      / Name of the REB
-CONTNUM =                      / CCD Controller (WREB) Serial Number
 TEMP_SET=                      / Temperature set point (deg C)
```
mareuter, line 94: This keyword and the next are in ComCam's sensor HDU, why not put them in LATISS's?

etc/LATISS/primary_hdu.header
```diff
@@ -91,24 +91,9 @@ IMAGETAG=                      / DAQ Image id
 OBSANNOT=                      / Observing annotation
 
         ---- Information from Camera ----
-CCD_MANU=                      / CCD Manufacturer
 CCD_TYPE=                      / CCD Model Number
```
mareuter: Is this supposed to be duplicated as its in the senor HDU file?

mareuter (changes requested):
Couple of questions about some keywords.

etc/LATISS/primary_hdu.header
menanteau: For LATISS, `TEMP_SET` comes a scalar from the telemetry in `ATCamera.fp_Ccd.temp`, so it stays on the common `PRIMARY` header template.

etc/LATISS/primary_hdu.header
menanteau: That was a duplicate and has been removed.

mareuter (approved):
Thanks for the feedback. Looks good now.

PR #93, opened 2021-06-02
Updates to ROTPA telemetry source

menanteau:


etc/conf/atTelemetry.yaml
```diff
@@ -333,26 +333,27 @@ telemetry:
...
+  array: undef
 
  RASTART:
   device: ATPtg
   Stype: Telemetry
   topic: mountPositions
   value: ra
   collect_after_event: start_collection_event
-  array: timestamp_start
+  array: undef
```
mareuter, line 348: I noticed a bunch of these changed. Before, was it automatically assuming the 0th element of the array and what you did now will allow one to select it?

mareuter (approved):
Looks fine with respect to the needed change. One question asked for behavior clarification.

etc/conf/atTelemetry.yaml
menanteau: I had plans to have a method called `timestamp_start` to extract from arrays, but it was never implemented. This was defaulting to select the first element from the array. I now changed to `undef` to make it clearer that it doing nothing special and hence picking up the first element of the array.

PR #100, opened 2022-04-14
Changes to use timeout parameter from Camera instead of exposure time.

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -456,11 +456,9 @@ async def complete_tasks_START(self, imageName):
...
 
             # Get the requested exposure time to estimate the total timeout
-            exptime_key = self.config.timeout_keyword
-            self.log.info("Collecting key to set timeout as key %s + %s" %
-                          (exptime_key, self.config.timeout_exptime))
-            metadata_tmp = self.collect([exptime_key])
-            timeout = metadata_tmp[exptime_key] + self.config.timeout_exptime
+            timeout_camera = self.read_timeout_from_camera()
+            timeout = timeout_camera + self.config.timeout_exptime
+            self.log.info(f"Setting timeout as camera_timeout + {self.config.timeout_exptime}")
```
mareuter: Be nice to report here what the camera timeout is actually being used.

mareuter (approved):
One small comment to consider, but otherwise this looks fine.

python/HeaderService/hslib_salobj.py
menanteau: Thanks for the suggestion.

PR #104, opened 2022-09-30
Changes for GenericCamera

menanteau:
Intermediate updates to Generic HS
Updates to work around missing metadata from GenericCamera.
Adding new code to create FITS file from yaml output
Updates to number of segments
Clean up of templates and making PRIMARY_SENSOR optional.

### DIFF
--- a/etc/GenericCamera-1/primary_hdu.header
+++ b/etc/GenericCamera-1/primary_hdu.header
@@ -1,0 +1,65 @@
+SIMPLE  =                    T /
+BITPIX  =                    8 / bits per data value
+NAXIS   =                    0 / number of axes
+EXTEND  =                    T / Extensions are permitted
+ORIGIN  = 'LSST DM Header Service'/ FITS file originator
+
+        ---- Date, night and basic image information ----
+TIMESYS = 'TAI'                / The time scale used
+DATE    =                      / Creation Date and Time of File
+DATE-OBS=                      / Date of observation (image acquisition)
+DATE-BEG=                      / Time at the start of integration
+DATE-END=                      / end date of the observation
+MJD     =                      / Modified Julian Date that the file was written
+MJD-OBS =                      / Modified Julian Date of observation
+MJD-BEG =                      / Modified Julian Date derived from DATE-BEG
+MJD-END =                      / Modified Julian Date derived from DATE-END
+OBSID   =                      / ImageName from Camera StartIntergration
+BUNIT   = 'adu     '           / Brightness units for pixel array
+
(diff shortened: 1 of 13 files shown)


python/HeaderService/__init__.py
```diff
@@ -20,7 +20,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 __author__ = "LSST/Felipe Menanteau"
-__version__ = "3.0.3"
```
mareuter, line 23: This should probably at least a minor version rev since we're adding a new camera system.

etc/GenericCamera-1/primary_hdu.header
```diff
@@ -0,0 +1,65 @@
...
+        ---- Additional Keys Information from Camera ----
+PROGRAM =                      / Name of the program
+REASON  =                      / Reason for observation
+
+        ---- Test Stand information ----
+TSTAND  =                      / The Test Stand used
+
+        ---- Information from Camera ----
+CAMMM   =                      / Camera Make and model
+DIAMETER=	                     / Lens diameter [mm]
```
mareuter: Any reason why the comment isn't at the same indent point as the others in this section?

python/HeaderService/hutils.py
```diff
@@ -422,15 +430,21 @@ def get_primary_extname(self, sensor):
 
     def get_segment_extname(self, sensor, seg):
         """Get the right SEGMENT extension name used for a sensor/CCD"""
-        extname = f"{sensor}_{self.segname}{seg}"
+        # Account for different formating for EXTNAME for LATISS
```
mareuter: I don't see a branch for LATISS here, so it this comment relevant?

mareuter (approved):
I've checked things over and the Confluence table matches what is in the code. I've got just a couple of minor comments. Also, looks like the lint job failed and you should remove the old Travis tests from the branch protections and set the GH action lint job for that role.

python/HeaderService/__init__.py
menanteau: Good point. I'll make it `3.1.0`

etc/GenericCamera-1/primary_hdu.header
menanteau: Thanks for catching that. There was a `tab` somewhere.

python/HeaderService/hutils.py
menanteau: Old comment. when that used to be the case for LATISS. Fixed.

PR #105, opened 2022-10-21
Changes for Star Tracker GenericCameras

menanteau:
Re-organization and updates to regular expression to define yaml
Add simulation script
Adding yaml files for 102 and 103


python/HeaderService/hsregex.py
```diff
@@ -0,0 +1,95 @@
...
+    for match in all_matches:
+        vars_dict[match] = d[match]
+    return vars_dict
+
+
+def update_var_dict(d, var_dict):
+    """
+    Updates the matches in dictionary with the new values
+    """
+    # Loop over the items    # Loop over the items
```
mareuter: I think you can remove the duplicate comment.

etc/conf/gc1Telemetry.yaml
```diff
@@ -195,11 +197,14 @@ telemetry:
...
   array: keyed_array
 
+# -- Additional Keywords from Camera
+
  REASON:
   device: GenericCamera
   Stype: Event
   topic: startIntegration
   value: additionalValues
+  device_index: 1
```
mareuter: Since you used the convention of `${hs_index}` in the other files, why not use that in this one as well.

etc/GenericCamera-101/primary_hdu.header
```diff
@@ -0,0 +1,75 @@
...
+MJD-OBS =                      / Modified Julian Date of observation
+MJD-BEG =                      / Modified Julian Date derived from DATE-BEG
+MJD-END =                      / Modified Julian Date derived from DATE-END
+OBSID   =                      / ImageName from Camera StartIntergration
+BUNIT   = 'adu     '           / Brightness units for pixel array
+
+        ---- Telescope info, location, observer ----
+FACILITY= 'Vera C. Rubin Observatory' / Facility name
+TELESCOP=                      / Telescope name
+INSTRUME= 'Simulation Camera'  / Instrument used
```
mareuter: Confluence page has this as `StarTracker`

etc/GenericCamera-101/primary_hdu.header
```diff
@@ -0,0 +1,75 @@
...
+MJD     =                      / Modified Julian Date that the file was written
+MJD-OBS =                      / Modified Julian Date of observation
+MJD-BEG =                      / Modified Julian Date derived from DATE-BEG
+MJD-END =                      / Modified Julian Date derived from DATE-END
+OBSID   =                      / ImageName from Camera StartIntergration
+BUNIT   = 'adu     '           / Brightness units for pixel array
+
+        ---- Telescope info, location, observer ----
+FACILITY= 'Vera C. Rubin Observatory' / Facility name
+TELESCOP=                      / Telescope name
```
mareuter: The Confluence page has been updated to fill in this keyword.

mareuter (commented):
Want to get a first set of comments out while I wait for the pointing information to get settled.

python/HeaderService/hsregex.py
menanteau: Thanks, fixed.

etc/GenericCamera-101/primary_hdu.header
menanteau: Good catch. Fixed.

etc/GenericCamera-101/primary_hdu.header
menanteau: Fixed too. Thanks for noticing.

etc/conf/gc1Telemetry.yaml
menanteau: Good point. Updated.

mareuter (approved):
Was able to test this on TTS and it looks like it's getting all the expected information. I've cross-verified with the Confluence table as well. I think this is ready to go.

PR #106, opened 2022-11-01
Adding expiration check to telemetry

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -727,6 +731,22 @@ def collect(self, keys):
                     self.log.warning(f"Cannot extract keyword: {keyword} from topic: {name}")
         return metadata
 
+    def check_telemetry_expired(self, myData):
+        """ Check is telemetry has expired using expiresAt parameter"""
+        has_expired = False
+        # Check if it has the expiresAt attribute
+        if hasattr(myData, 'expiresAt'):
+            expiresAt = getattr(myData, 'expiresAt')
+            expiresIn = getattr(myData, 'expiresIn')
```
mareuter: Is the extraction of this parameter and the one on the next line for future use? If not, I think they should be removed since they are not used in the expiration check. Same goes for their corresponding log messages below.

menanteau: Yeah, it makes sense. I will clean it up.

PR #107, opened 2022-11-09
Update telemetry source for CCDTEMP on AuxTel

menanteau:


bin/telemetry_sim_LATISS
```diff
@@ -326,7 +326,8 @@ async def amain():
         await cam.tel_focal_plane_Ccd.set_write(**kwCCD)
 
         # Send CCD temperature Setpoint telemetry using a random generator
-        kwCCD = {'tempCCDSetPoint': random.uniform(-50, -40)}
+        kwCCD = {'tempCCDSetPoint': random.uniform(-50, -40),
```
patrickingraham: Note that the steady state value is about -94 degrees Celsius, so I'd put the ranges from -100 to -90 to be more realistic.

patrickingraham (approved):
Made a single comment about the simulated values.

bin/telemetry_sim_LATISS
menanteau: Thanks for suggestion. It's fixed.

PR #109, opened 2022-12-16
Update to go to Fault state when it cannot write headers to S3

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -671,11 +683,25 @@ async def announce(self, imageName):
...
+        return s3upload
 
     def write(self, imageName):
         """ Function to call to write the header"""
         self.HDR[imageName].write_header(self.filename_HDR[imageName])
-        self.log.info("Wrote header to: {}".format(self.filename_HDR[imageName]))
+        self.log.info("Wrote header to filesystem: {}".format(self.filename_HDR[imageName]))
 
     def clean(self, imageName):
         """ Clean up imageName data structures"""
```
ktlim, line 709: Do you need to `del self.completed_OK[imageName]` here so that it doesn't become a memory leak?

python/HeaderService/hslib_salobj.py
```diff
@@ -671,11 +683,25 @@ async def announce(self, imageName):
...
+        s3upload = False
+        k = 1
+        while k <= nt and s3upload is False:
+            try:
+                await self.s3bucket.upload(fileobj=fileobj, key=key)
+                s3upload = True
+            except Exception as e:
+                self.log.error(f"Failed s3bucket.upload attempt # {k}/{nt} for: {imageName}")
+                self.log.exception(str(e))
+            k += 1
```
ktlim, line 700: Maybe do a `time.sleep(0.1)` before retrying?  (Well, ideally make the delay a module constant or configuration value rather than hard-coding it.  You could also make the number of retries a constant or configuration value.)

python/HeaderService/hslib_salobj.py
```diff
@@ -671,11 +683,25 @@ async def announce(self, imageName):
 
         await self.evt_largeFileObjectAvailable.set_write(**kw)
         self.log.info(f"Sent {self.config.hs_name} largeFileObjectAvailable: {kw}")
+        self.completed_OK[imageName] = True
+
+    async def upload_to_s3(self, fileobj, key, imageName, nt=2):
```
ktlim, line 689: This is fairly self-explanatory, but a minimal docstring would still be welcome.

ktlim (approved):

python/HeaderService/hslib_salobj.py
menanteau: Yep. Thanks for noticing.

python/HeaderService/hslib_salobj.py
menanteau: It's not needed. The internal `s3bucket.upload` has a long wait built in.

python/HeaderService/hslib_salobj.py
menanteau: Will do.

PR #110, opened 2023-01-19
Adding Weather telemetry to ComCam to be like LATISS

menanteau:


bin/telemetry_sim_ComCam
```diff
@@ -125,6 +140,8 @@ async def amain():
     dome = salobj.Controller(name="MTDome", index=0)
     rotator = salobj.Controller(name="MTRotator", index=0)
     dometraj = salobj.Controller(name="MTDomeTrajectory", index=0)
+    dimm = salobj.Controller(name="DIMM", index=1)
```
ktlim, line 143: Just checking: we don't have separate DIMM/WeatherStation information (at a different index) for the Simonyi Telescope versus the AuxTel?

ktlim (approved):

bin/telemetry_sim_ComCam
menanteau: Will double check with folks.

PR #114, opened 2023-03-15
Adding exception handlers to collect method.

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -757,8 +757,16 @@ def collect(self, keys):
                     if 'scale' in self.config.telemetry[keyword]:
                         metadata[keyword] = metadata[keyword]*self.config.telemetry[keyword]['scale']
                         self.log.info(f"Scaled key: {keyword} by: {self.config.telemetry[keyword]['scale']}")
-                except KeyError:
-                    self.log.warning(f"Cannot extract keyword: {keyword} from topic: {name}")
+                except ValueError as err:
+                    self.log.error(f"Cannot extract keyword: {keyword} from topic: {name}")
```
tribeiro: Doesn't seem like there anything different in the error handling for any of the exceptions you are catching.. Consider doing something like this:

```
except Exception:
    self.log.exception(f"Cannot extract keyword: {keyword} from topic: {name}")
```

This will log the original traceback alongside the message provided, so you don't need to `self.log.error(f"Exception: {err}")`.

tribeiro (approved):
I left a comment in line I hope you will consider before merging. 

Thanks for the update.

python/HeaderService/hslib_salobj.py
menanteau: Agreed. That was my original ideal, but ended up overkilling it.

PR #117, opened 2023-04-26
Updates to GenericCamera HS with new Camera Info.

menanteau:


mareuter:
You need to update `gc102Telemetry.yaml` and `gc103Telemetry.yaml` with the same fixes as `gc101Telemetry.yaml`.

menanteau:
Thanks for catching this. I was under the impression that there were all the same (i.e. linked) but that's only the case for the header templates.

mareuter (approved):
Looks good to me.

PR #120, opened 2023-05-04
Updates to handle_summary_state fuction

menanteau:
Updates to handle_summary_state

### DIFF
--- a/python/HeaderService/hslib_salobj.py
+++ b/python/HeaderService/hslib_salobj.py
@@ -39,6 +39,9 @@ class HSWorker(salobj.BaseCsc):
 
     def __init__(self, **keys):
 
+        # initialize current state to None
+        self.current_state = None
+
         # Load the configurarion
         self.config = types.SimpleNamespace(**keys)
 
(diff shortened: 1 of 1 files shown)


python/HeaderService/hslib_salobj.py
```diff
@@ -397,6 +415,18 @@ def check_outdir(self, filepath):
             os.makedirs(filepath)
             self.log.info(f"Created dirname:{filepath}")
 
+    def check_services(self):
+        """ Check that services needed s3/web are working"""
+
+        self.log.info("Checking for services")
+        # Define/check s3 buckets
+        if self.config.lfa_mode == 's3':
+            self.define_s3bucket()
```
srp3rd: If this line fails, it throws an exception - instead of that, do you want to call self.fault() instead?

srp3rd: Also check_services() is called from prepare() on line 52, which if s3 fails will cause an exception.  You probably want to remove check_services() from prepare(), since you all it on line 122.

srp3rd (changes requested):
Additionally, the end_evt_timeout_task list isn't cleared after cancel_timeout_tasks() is called.  If this is called again later, all of the previously canceled tasks are still in the list, and cancel() will be called on each of those again.  I'm not sure what will happen if you call cancel() twice on the same task.

python/HeaderService/hslib_salobj.py
menanteau: Thanks. I changed the code as you suggested.

PR #123, opened 2024-02-07
Update to read telemetry from external file.

menanteau:
Update to CC Telemetry configuration and order of updates.
Adding more json files to library


mareuter:
I don't see any indication that the HS is in simulation mode with this. The simulation mode construct that salobj provides handles this for you. https://ts-salobj.lsst.io/salobj_cscs.html#simulation-mode

menanteau:
> I don't see any indication that the HS is in simulation mode with this. The simulation mode construct that salobj provides handles this for you. https://ts-salobj.lsst.io/salobj_cscs.html#simulation-mode

Updated to include `hs_simulation_mode` as a configurable option

mareuter:
The logging is nice, but you aren't sending out `logevent_simulationMode` with the `mode` attribute set to a non-zero number. That's the requirement that's missing.

menanteau:
> The logging is nice, but you aren't sending out `logevent_simulationMode` with the `mode` attribute set to a non-zero number. That's the requirement that's missing.

Added `simulationMode` event at startup.

mareuter (approved):
Looks good to me.

PR #124, opened 2024-02-09
Update to go to FAULT when it cannot update metadata from json file.

menanteau:
Fix bug that prevented update of values from json file.


mareuter (approved):
Looks good to me.

PR #126, opened 2024-02-09
Adding more telemetry keywords to keep, and making them configurable

menanteau:


etc/conf/ccTelemetry.yaml
```diff
@@ -709,3 +709,15 @@ playback_keywords:
  - HIERARCH SIMULATED MJD-OBS
  - HIERARCH SIMULATED DATE-OBS
  - HIERARCH SIMULATED DATE-END
+
+# Keywords not to ignore from telemetry in playback mode
+playback_keywords_keep:
+  - EMUIMAGE
+  - 'DATE-OBS'
```
mareuter: Are the single quotes here for a special reason? The `HIERARCH` lines above don't have quotes and they contain words with dashes.

mareuter (approved):
One minor comment, but otherwise this looks fine.

etc/conf/ccTelemetry.yaml
menanteau: I was worried that the `-` was going to mess up the string. But now I know that it wont't.

PR #142, opened 2024-10-18
Update FOCUSZ units and scale from microns to mm for ComCam and LSSTCam

menanteau:


ktlim (approved):
Approved as to syntax.  Someone else needs to review for content.

gmegh (approved):
LGTM!

PR #150, opened 2025-03-07
Updates to logging to start HeaderService using kafka only.

menanteau:
Adding await to CSCs
More updates for Cycle 40+ Kafka versions


mareuter (approved):
Looks good to me.

PR #151, opened 2025-03-11
Adding functionality to collect metadata at custom event from Camera.

menanteau:
Updates for HeaderService to collect from custom events from Camera.


mareuter (approved):
Looks good to me. We should document in the Confluence pages what metadata has the  variations in the collection events from the standard ones.

PR #153, opened 2025-04-10
Adding separate connection to mageParam_event, this will fix LSSTCam --playback

menanteau:
Adding separate connection to mageParam_event, this will fix LSSTCam --playback


mareuter (approved):
Looks good to me.

mfisherlevine (approved):

PR #155, opened 2025-04-11
Update yaml configuration files for GCHeaderService

menanteau:
Update yaml configuration files for GCHeaderService to include cameraConf_event parameter


mareuter (approved):
Looks good to me.

PR #158, opened 2025-04-24
Adding ROTSTART and ROTEND to LSSTCam telemetry.

menanteau:


etc/LSSTCam/primary_hdu.header
```diff
@@ -40,6 +40,8 @@ RAEND   =                      / RA of telescope from AZEND and ELEND
 DECEND  =                      / DEC of telescope from AZEND and ELEND
 ROTPA   =                      / Rotation angle relative to the sky (deg)
 ROTCOORD= 'sky'                / Telescope Rotation Coordinates
+ROTSTART=                      / Physical rotator position at start (deg)
+ROTEND  =                      / Physical rotator position at end (deg)
```
timj:
```suggestion
ROTEND  =                      / [deg] Physical rotator position at end
```
and the other one (and I see ROTPA has the unit in the wrong place in the string as well.

etc/LSSTCam/primary_hdu.header
```diff
@@ -40,6 +40,8 @@ RAEND   =                      / RA of telescope from AZEND and ELEND
...
 HASTART =                      / [HH:MM:SS] Telescope hour angle at start
 ELSTART =                      / [deg] Telescope zenith distance at start
```
timj: This comment is wrong. Elevation is not zenith distance.

menanteau: What do you want the comment to be Elevation?

etc/LSSTCam/primary_hdu.header
menanteau: Sure, can do.

etc/LSSTCam/primary_hdu.header
timj: yes. `[deg] Telescope elevation at start`

(and ELEND presumably).

menanteau: Got it. Thanks.

mareuter (approved):
When Tim is happy with this PR, then I will be too.

menanteau:
Are you satisfied with changes @timj ?

timj (approved):
Thanks. Ok with me.

PR #159, opened 2025-04-29
Update topics for each device when creating Remotes.

menanteau:


python/HeaderService/hslib_salobj.py
```diff
@@ -1390,6 +1422,10 @@ def extract_telemetry_channels(telem, start_collection_event=None,
         # Make sure we don't create extra channels
         if name not in channels.keys():
             channels[name] = telem[key]
+        # Store the topics for the device
+        if collect_dict is not None:
+            device_topics = collect_device_topics(collect_dict, device_topics)
+        device_topics = collect_device_topics(telem[key], device_topics)
```
mareuter, line 1428: If `collect_dict` is not None then `device_topics` is already set and then redone here. Is that what is intended?

mareuter (approved):
One item to look at but otherwise this looks good.

python/HeaderService/hslib_salobj.py
menanteau: `collect_dict` in `None` in case `collect_after_event` in not a dict  but instead a `str` in the set `FIXED_COLLECTION_EVENTS`

PR #162, opened 2026-04-02
Added Calibration header keywords to primary header hdu

mxturri:


tony-johnson (approved):
Looks good to me. We should be able to test that we see these entries being published at BTS, even if they are empty.

PR #163, opened 2026-06-15
Updated LED_NAMES/LEVELS to listen to new ledsOn events.

mxturri:


etc/LSSTCam/primary_hdu.header
```diff
@@ -117,5 +117,5 @@ HIERARCH SIMULATED DATE-OBS= /
 HIERARCH SIMULATED DATE-END= /
 HIERARCH CALIB LASER_WAVE= / Calibration Laser Wavelength
 HIERARCH CALIB LASER_MODE= / Calibration Laser Mode
-HIERARCH CALIB LED_NAMES= / Calibration LED Name
-HIERARCH CALIB LED_LEVEL= / Calibration LED Level
\ No newline at end of file
+HIERARCH CALIB LED_NAMES= / Calibration LED Names
+HIERARCH CALIB LED_LEVELS= / Calibration LED Levels
```
timj: Reminder that changing the header names requires we bump the global file header version so we should only do this if we think it's really important (since we have written files with the old name)

cc/ @tony-johnson

mxturri: @timj I made that change because I _thought_ you would want it, to be consistent with the plural in LED_NAMES.
If you don't think this is really important, I am happy to put it back as it was.
Please advise.

timj: It is a better name but if we change the name then code has to be modified to work with both and we have to document it as a distinct header variant so it's more trouble than it's worth. You can fix the comment strings. What I don't understand is how you are going to represent multiple values in the header.

Are we still event based rather than the CSC knowing the values itself and publishing them directly?

mxturri: @couger01 has added a new LEDProjector event called "ledsOn" with two fields:
- serialNumbers: a String with comma separated names of the leds that are on
- dacValues: a String with the comma separated dac values for the corresponding leds that are on

The Header service is passing those values through without any manipulation.

I will remove the S in LED_LEVELS.

tony-johnson: @timj Note that this is more than a name change. Previously HIERARCH CALIB LED_LEVEL was published to FITS as a Float (on the mistaken assumption there was only one LED/LEVEL). Now both LED_NAMES and LED_LEVELS will be publish as comma delimited strings (even if there is only one LED). This will require a FITS version # increment (and a separate JIRA).

And to be honest I am not sure what that will look like at the FITS level, I am guessing "1.0 , 0.9"

timj: In that case we may as well fix the name as well because you have to write special code for the new files anyhow...

Re: events. I still don't like events vs CSC state since the header service missing an event leads to the wrong answer whereas the header service retrieving the actual state is always going to be correct.

mxturri: I have added the "S" at the end of LED_LEVEL.

couger01: I'm not sure if I have heard of an event being missed by a remote vs an event not being published as expected.

